PR #26, opened 2017-08-14
Feature/clean code improvements

### LINKED ISSUE #19: Use meaningful variables to avoid 'mental mapping'
While this project does not have a clearly defined coding standard, it is a good start to make sure we use meaningful variables to avoid mental mapping. This means that all variables should explain what they are, rather than using, for example, placeholders.

```py
# bad
a = get_apple()
o = get_orange()
s = make_smoothie(a, o)

# good
apple =get_apple()
orange = get_orange()
smoothie = make_smoothie()
```

# Resources
Clean code JavaScript has some good examples that equally apply to Pythin:
- [Use meaningful and pronounceable variable names](https://github.com/ryanmcdermott/clean-code-javascript#use-meaningful-and-pronounceable-variable-names)
- [Use searchable names](https://github.com/ryanmcdermott/clean-code-javascript#use-searchable-names)
- [Avoid Mental Mapping](https://github.com/ryanmcdermott/clean-code-javascript#avoid-mental-mapping)

brylie:
Closes #19 

# Changes
- use explanatory/searchable variables where possible

# Scope of changes
Perhaps this PR should be limited to the `apis` app.


brylie:
I labeled this as 'ready for review' so we can limit the scope. Other changes can take place in subsequent PRs.

brylie:
Yeah, we should rename the variable. Double check that it has not been renamed in a different PR. What would be a more meaningful name?

tractaylor:
project/api/models/civi.py - lines 123  -- this looks like a created date, so maybe created_date or date_created?

project/api/models/civi.py - lines 146 -- not really sure about this one.  Its always 1 or 0, and I really can't tell what is doing besides just adding that to the ranking score on line 179, 186, 194, 196

PR #56, opened 2017-09-04
Draft Key Functionality report for reference and edit

brylie:
Add key functionality report. License documentation as Creative Commons Zero.


brylie:
@mdwest I moved the Key Functionality report into the correct folder. I also added the Creative Commons Zero license in the `/docs` folder, to indicate that all documentation is Public Domain (worldwide jurisdiction).

PR #85, opened 2017-09-25
Open Canvas V1.0

mdwest:


brylie:
Perhaps a different folder than `design` would be better here. E.g. `organization`, for documents related to the organization vision, meetings, etc.?

mdwest:
@brylie I think I'm making a mess trying to move this (Almost added a second pull request). Lets talk Monday to get this moved to the correct folder.

PR #110, opened 2017-12-18
Reorganize imports

### LINKED ISSUE #41: Reorganize imports to follow best practices
Many files in this project have a 'mish-mash' import style, where imports seem to have been added through accretion.

Two Scoops of Django (1.11) recommends that projects adopt the following import ordering:

1. Standard Library imports
2. Django core imports
3. third-party app imports
4. project internal imports

# Goal
Use consistent import ordering throughout project.

# Task
Re-organize imports in the following files:

- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/dev/api/models/civi.py#L1-L12
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/dev/api/write.py
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/dev/api/tasks.py
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/dev/api/read.py
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/dev/api/models/invitation.py#L4-L7
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/dev/api/models/thread.py
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/dev/authentication/authentication.py#L1-L24
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/dev/authentication/forms.py#L1-L15
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/dev/frontend_views/views.py#L1-L14

jl24:
Resolves #41

### DIFF
--- a/project/api/write.py
+++ b/project/api/write.py
@@ -36,6 +36,10 @@ def new_thread(request):
     if state:
         new_thread_data['state'] = state
 
+    is_draft = request.POST['is_draft']
+    if is_draft:
+        new_thread_data['is_draft'] = is_draft
+
     new_t = Thread(**new_thread_data)
     new_t.save()
 
--- a/project/api/write.py
+++ b/project/api/write.py
@@ -11,7 +11,6 @@ from django.contrib.auth.decorators import login_required
 from django.contrib.auth.decorators import user_passes_test
 from django.contrib.sites.shortcuts import get_current_site
 
-
 # civi packages
 from api.forms import UpdateProfileImage
 from api.models import Thread
@@ -19,8 +18,6 @@ from api.tasks import send_mass_email
 from models import Account, Activity, Category, Civi, CiviImage, Invitation
 from utils.custom_decorators import require_post_params
 from utils.constants import US_STATES
-from utils.custom_decorators import require_post_params
-
 
 @login_required
 @require_post_params(params=['title', 'summary', 'category_id'])
--- a/project/api/write.py
+++ b/project/api/write.py
@@ -1,26 +1,20 @@
 import json, PIL, urllib, uuid
 
-from notifications.signals import notify
-
-# django packages
-from django.contrib.auth.models import User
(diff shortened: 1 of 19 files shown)


project/api/write.py
```diff
@@ -254,46 +251,63 @@ def deleteCivi(request):
...
+
+    if not is_draft:
+        try:
+            draft_thread = Thread.objects.get(id=thread_id)
+        except Exception as e:
+            return HttpResponseServerError(reason=str(e))
+
+        draft_thread.is_draft = False
+
+        # try:
```
brylie: Clean up commented code.

project/api/write.py
```diff
@@ -254,46 +251,63 @@ def deleteCivi(request):
...
-            },
-            "level": t.level,
-            "state": t.state if t.level == "state" else "",
-            "location": t.level if not t.state else dict(US_STATES).get(t.state),
-        }
-        return JsonResponse({'data': return_data})
-    else:
-        return HttpResponseBadRequest(reason="Invalid Thread Reference")
+    try:
+        t = Thread.objects.get(id=thread_id)
```
brylie: Use meaningful variable name, e.g. `thread`.

project/api/write.py
```diff
@@ -254,46 +251,63 @@ def deleteCivi(request):
...
+        t.summary = summary
+        t.category_id = category_id
+        t.level = level
+        t.state = state
+        t.save()
+    except Exception as e:
+        return HttpResponseServerError(reason=str(e))
 
+    return_data = {
+        'thread_id': thread_id,
```
brylie: Can we access the ID property directly from the `thread` object? If so, it would be consistent to do something like `thread.id` here, rather than using the variable `thread_id`. The remainder of this object uses properties of `thread` instead of placeholder variables.

jl24:
@brylie I think i messed up the commits between branches and got some other changes mixed up here. This branch should only have reorganizing of the imports but seems to have more. Going to look into it

PR #129, opened 2018-02-12
Adding the test framework

### LINKED ISSUE #118: Initial testing effort
In order to have confidence when releasing new versions of CiviWiki, it would be good to have some automated test coverage. This issue is to track progress towards the initial test coverage plan.

# Resources
See the [Initial Test Plan](https://github.com/CiviWiki/OpenCiviWiki/wiki/Initial-Test-Plan) document for more details.

Unit111:
closes #118

The test_framework dir contains a venv with all the necessary packages for running it.
If you prefer using  global packages, you only need to install two:
aloe (0.1.15)
selenium (3.8.0)
The versions are the ones I use. There should be no problem using newer.
The tests themselves can be run using the "aloe" command. My advice is running in from inside the project directory, for example:
aloe E:\Users\Unit 1\PycharmProjects\CiviWikiTest
running it on any other level will cause errors.
If you prefer running individual features you can do it in a similar way:
aloe <path_to_feature_file> 

Scenario: User can create threads is unfinished since there is no delete thread functionality implemented.
It currently only creates new threads with the same name.


brylie:
Lets add `.idea` to `.gitignore`, so we don't keep that folder under revision control.

project/test_framework/.idea/CiviWikiTest.iml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
brylie: Lets add `.idea` to `.gitignore`, so we don't keep that folder under revision control.

project/test_framework/venv/pyvenv.cfg
```diff
@@ -0,0 +1,3 @@
+home = E:\Program Files\Python\Python36-32
```
brylie: I would prefer not to keep the venv files under revision control. Rather, we should see if we can keep a virtual environment description file, such as a [conda](https://conda.io/docs/) `environment.yml` or a [pipenv](https://github.com/pypa/pipenv) file. That way, all dependencies are downloaded from external source, rather than kept in our repository.

petarPetrovBG:
Comments applied.

PR #163, opened 2018-03-11
Csv export pull request

### LINKED ISSUE #148: Create CSV export option for Civi data
Allow Civis to be exported via CSV. This will allow for analysis and integration with third party tools.

Parent issue: #101

delta-plus:
Closes #148, **CSV export option for civi data**. This adds a button to civi threads, next to the "Add Civi" button. The button allows all of the text in each civi card in the thread to be organized into a .csv file which is immediately presented to the user for download without leaving the page. It's done with an Ajax call to the function civi2csv in views.py.

There are a few requirements for it to work properly:
- Python libraries django-cors-headers, bs4, and csv must be installed.
-The (webroot)/temp directory must be available for writing .csv files to.

If there are any problems or you want more information about the feature, you can message me on the riot chat or at j03grassl@gmail.com.

### DIFF
--- a/project/civiwiki/settings.py
+++ b/project/civiwiki/settings.py
@@ -51,10 +51,12 @@ INSTALLED_APPS = (
     'frontend_views',
     'notifications',
     'legislation',
+    'corsheaders',
 )
 
 
 MIDDLEWARE_CLASSES = (
+    'corsheaders.middleware.CorsMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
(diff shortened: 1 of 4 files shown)


brylie:
Please add any dependencies to the project [requirements.txt](https://github.com/CiviWiki/OpenCiviWiki/blob/dev/requirements.txt).

delta-plus:
I hardcoded localhost because that's the address that worked for me. As for
the civis being in the database, I didn't know where to access it, so I
just passed the HTML as JSON and got the text from it. It works and it's
fast, but the options you suggest do seem better overall. I'll redo the
function and document the new approach.

brylie:
Querying the civis directly from the database will take a bit more learning, but will leave us on better footing for the long term. It will also help you gain familiarity with the CiviWiki data model.

Please let us know if/when you encounter difficulties figuring things out :-)

brylie:
I just want to say that this is a clever solution! :-)

Lets just stick a bit closer to Django conventions, particularly those outlined in the [official tutorial](https://docs.djangoproject.com/en/2.0/intro/tutorial01/). Specifically, section 2 shows [how to use the Django shell to interact with internal APIs](https://docs.djangoproject.com/en/2.0/intro/tutorial02/#playing-with-the-api).

PR #164, opened 2018-03-19
Django REST Framework

### LINKED ISSUE #24: Use Django REST Framework for REST endpoints
There is currently a TODO in our `api/urls.py` to build the API using Django REST Framework (DRF). This issue is to earmark that work, and track progress towards the goal of using DRF for CiviWiki REST APIs.

### LINKED ISSUE #120: REST API endpoint(s) for threads
As part of our work to port the CiviWiki APIs to use Django REST Framework, we can create a full-fledged CRUD API for Threads. This API should stick close to the DRF documentation, including possibly using [ModelSerializer](http://www.django-rest-framework.org/api-guide/serializers/#modelserializer) and other design conventions.

If possible, the Threads and thread related functionality (including the REST API) should be contained in their own app.

# Original request
I want to add an automation test for creating threads. But in order to avoid spamming new threads on every automation run I need to be able to delete the ones created by the automation.

jl24:
Closes #24
Closes #120 

Incorporates the Django REST Framework into our api Django app.


project/api/serializers.py
```diff
@@ -0,0 +1,84 @@
+from rest_framework import serializers
+from api.models import Civi, Thread, Account, Category
+
+
+WRITE_ONLY = {'write_only': True}
+
+class AccountSerializer(serializers.ModelSerializer):
```
brylie: We need to consider what fields of the user profile should be publicly available. I.e. we need to think about user privacy here. Information like first/last name, address/location, and picture are personal, so should be treated with care.

brylie:
Looks good for the most part.

My main concern right now is about user privacy. I.e. we need to be intentional and communicative about what personal information we are making available.

jl24:
@brylie  that was one of my main concerns too. With the whole social media privacy issue going on, how much user info do we want to have publicly available? We want an open system that promotes collaboration and freely accessible information in terms of the discussion, and user identity is something we don't want to totally stifle either. Perhaps we should think about incorporating a privacy setting that will determine how much public facing info a user can have, so that we give the users a choice upon the matter.

jl24:
Some Notes:
- The browsable API that comes with DRF is enabled with the environment variable `DEBUG` set to `True` 
- There is an auto-generated api under /api/docs which gives a general overview. We can go to different 3rd party documentation libraries if needed.
- I've followed the` /api/v1/...` url format that was present in the bills PR

I've also addressed few of the the security concerns:
1. The full listing of all user accounts is limited to admin users with staff access only. 
2. User Account query is done through the username instead of the id value (since our id is a simple sequence right now, which maybe another point of discussion)
3. Any account related querying requires authentication.

PR #169, opened 2018-04-02
Csv export pr update

### LINKED ISSUE #148: Create CSV export option for Civi data

delta-plus:
Closes #148

This is the _new_, new and improved CSV export feature.

**How it Works**
An entry in frontend_views/urls.py ensures that any URL that looks like /thread/[thread ID]/csv triggers the civi2csv view in frontend_views/views.py. An HTTP response is returned to the user containing the CSV data as an attachment, which is presented for download automatically. The export button is added to webapp/templates/partials/thread/thread_body.html and has a dynamic href parameter thanks to an added "thread_id" entry in the thread_wiki_data dict of the issue_thread view.

**Notes**
Should be good to go!
If the CSV files should be formatted differently, please let me know.

@jl24 @brylie

### DIFF
--- a/project/civiwiki/settings.py
+++ b/project/civiwiki/settings.py
@@ -169,15 +169,22 @@ STATIC_ROOT = os.path.join(BASE_DIR, 'staticfiles')
 
 
 # Database
-DATABASES = {
-    'default': {
-        'HOST': 'localhost',
-        'PORT': '5432',
-        'NAME': 'postgres',
-        'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'USER': 'postgres',
-        'PASSWORD': 'asdf',
-    },
+if 'CIVIWIKI_LOCAL_NAME' not in os.environ:
+    STATICFILES_STORAGE = 'whitenoise.django.GzipManifestStaticFilesStorage'
+
+    DATABASES = {
+        'default': dj_database_url.parse(get_env_variable("DATABASE_URL"))
+    }
+else:
+    DATABASES = {
+        'default': {
+            'HOST': 'localhost',
+            'PORT': '5432',
+            'NAME': get_env_variable("CIVIWIKI_LOCAL_NAME"),
+            'ENGINE': 'django.db.backends.postgresql_psycopg2',
+            'USER': get_env_variable("CIVIWIKI_LOCAL_USERNAME"),
+            'PASSWORD': get_env_variable("CIVIWIKI_LOCAL_PASSWORD"),
+        },
 }
 
 
--- a/project/civiwiki/settings.py
+++ b/project/civiwiki/settings.py
@@ -185,7 +185,7 @@ else:
             'USER': get_env_variable("CIVIWIKI_LOCAL_USERNAME"),
             'PASSWORD': get_env_variable("CIVIWIKI_LOCAL_PASSWORD"),
         },
-}
+    }
 
 
 # Email Backend Setup
--- a/project/civiwiki/settings.py
+++ b/project/civiwiki/settings.py
@@ -51,10 +51,12 @@ INSTALLED_APPS = (
     'frontend_views',
     'notifications',
     'legislation',
+    'corsheaders',
 )
 
 
 MIDDLEWARE_CLASSES = (
+    'corsheaders.middleware.CorsMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
@@ -167,23 +169,16 @@ STATIC_ROOT = os.path.join(BASE_DIR, 'staticfiles')
 
 
 # Database
-if 'CIVIWIKI_LOCAL_NAME' not in os.environ:
-    STATICFILES_STORAGE = 'whitenoise.django.GzipManifestStaticFilesStorage'
-
-    DATABASES = {
-        'default': dj_database_url.parse(get_env_variable("DATABASE_URL"))
-    }
-else:
-    DATABASES = {
-        'default': {
-            'HOST': 'localhost',
-            'PORT': '5432',
-            'NAME': get_env_variable("CIVIWIKI_LOCAL_NAME"),
-            'ENGINE': 'django.db.backends.postgresql_psycopg2',
-            'USER': get_env_variable("CIVIWIKI_LOCAL_USERNAME"),
-            'PASSWORD': get_env_variable("CIVIWIKI_LOCAL_PASSWORD"),
-        },
-    }
+DATABASES = {
+    'default': {
+        'HOST': 'localhost',
+        'PORT': '5432',
+        'NAME': 'postgres',
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
+        'USER': 'postgres',
+        'PASSWORD': 'asdf',
+    },
+}
 
 
 # Email Backend Setup
@@ -203,3 +198,6 @@ else:
 # Notification API Settings
 NOTIFICATIONS_SOFT_DELETE = True
 NOTIFICATIONS_USE_JSONFIELD = True
+
+# CORS Settings
+CORS_ORIGIN_ALLOW_ALL = True
--- a/project/frontend_views/urls.py
+++ b/project/frontend_views/urls.py
@@ -16,5 +16,5 @@ urlpatterns = [
     url(r'^invite$', v.invite, name='invite'),
     url(r'^beta_register/(?P<email>[\w.%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,4})/(?P<token>\w{31})$', v.beta_register, name='beta_register'),
     url(r'^$', v.base_view, name='base'),
-    url(r'^civi2csv$', v.civi2csv, name='civi2csv'),
+    url(r'^thread/(?P<thread_id>\w+)/csv$', v.civi2csv, name='civi2csv'),
 ]
--- a/project/frontend_views/urls.py
+++ b/project/frontend_views/urls.py
@@ -16,4 +16,5 @@ urlpatterns = [
     url(r'^invite$', v.invite, name='invite'),
     url(r'^beta_register/(?P<email>[\w.%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,4})/(?P<token>\w{31})$', v.beta_register, name='beta_register'),
     url(r'^$', v.base_view, name='base'),
+    url(r'^civi2csv$', v.civi2csv, name='civi2csv'),
 ]
--- a/project/frontend_views/views.py
+++ b/project/frontend_views/views.py
@@ -104,6 +104,7 @@ def issue_thread(request, thread_id=None):
     t.refresh_from_db()
 
     thread_wiki_data = {
+        "thread_id": thread_id,
         "title": t.title,
         "summary": t.summary,
         "image": t.image_url,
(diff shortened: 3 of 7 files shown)


project/webapp/templates/thread.html
```diff
@@ -6,6 +6,24 @@
     {% include "base/less_headers/thread_less.html" %}
 </head>
 
+<script>
+$(document).ready(function() {
+  $('#csv-export').click(function() {
+    $.ajax({
+      type: 'POST',
+      url: 'http://127.0.0.1:8000/civi2csv',
```
brylie: We have to be careful about tying code too closely to our development environment. `localhost:8000` will work for you when working on your development computer, but will break when the site is deployed. So, change this URL to a relative path, rather than hard-coding it to the development environment.

project/webapp/templates/thread.html
```diff
@@ -6,6 +6,24 @@
...
+      data: JSON.stringify({'thread': /[^/]*$/.exec(window.location.href)[0]})
```
brylie: What does this line do after 'thread'? Particularly, why do we need to use `exec(...`?

project/civiwiki/settings.py
```diff
@@ -167,23 +169,16 @@ def get_env_variable(environment_variable, optional=False):
...
-        'default': {
-            'HOST': 'localhost',
-            'PORT': '5432',
-            'NAME': get_env_variable("CIVIWIKI_LOCAL_NAME"),
-            'ENGINE': 'django.db.backends.postgresql_psycopg2',
-            'USER': get_env_variable("CIVIWIKI_LOCAL_USERNAME"),
-            'PASSWORD': get_env_variable("CIVIWIKI_LOCAL_PASSWORD"),
-        },
-    }
+DATABASES = {
```
brylie: Use the **environment variables** to store your development database configuration. That way this code stays generic to development and production environments.

Also, be careful about removing the STATICFILES_STORAGE variable, as it may be important for production.

delta-plus:
1. Yeah, I just wasn't sure what path you wanted. Would just "civi2csv" or
"/civi2csv" work?

2. That line is just some regex to chop the thread ID off the end of the
current URL.

3. Oh, right. Totally forgot about the database settings. I'll set those
back to defaults.

Thanks a bunch!

project/frontend_views/views.py
```diff
@@ -266,3 +267,20 @@ def about_view(request):
 
 def support_us_view(request):
     return TemplateResponse(request, 'static_templates/support_us.html', {})
+
+@csrf_exempt
+def civi2csv(request):
+    import csv
+    if request.method == 'POST':
```
brylie: Try removing this line to see if it simplifies the client code. All we should need on the client with an href, which makes a GET request. Checking for POST may be over-complicating this solution.

project/webapp/templates/thread.html
```diff
@@ -6,6 +6,24 @@
     {% include "base/less_headers/thread_less.html" %}
 </head>
 
+<script>
+$(document).ready(function() {
```
brylie: This script should be unnecessary, if you simply change the backend logic to not check for a POST request.

project/webapp/templates/partials/thread/thread_body.html
```diff
@@ -36,9 +36,12 @@
                             <span class="text">Add Civi</span>
                         </button>
                     </div>
+                    <div class="add-button" style="margin-left: 4px">
+                        <button id="csv-export" class="btn waves-effect">
```
brylie: Change this to an anchor tag with a *relative* HREF.

brylie:
Regarding the url path, lets just add a `/csv` postfix to match the existing Civi route. That way it is consistent and self documenting.

delta-plus:
Okay, but how would I get the response data? With Ajax or XMLHttpRequest I
could get the response pretty easily. Using GET is fine, but I'm not sure
how I would do this without a script.

brylie:
> Okay, but how would I get the response data? 

It should work just like requesting any other page or resource on the www. We just click a link and the browser males a request and handles the response. The content type (CSV) header tells the browser to treat the response as a file, so it should show a download confirmation.

delta-plus:
Okay, I could create the download href automatically as the thread loads
using a view or something. That could work.

On Apr 4, 2018 11:58 PM, "Brylie Christopher Oxley" <
notifications@github.com> wrote:

Okay, but how would I get the response data?

It should work just like requesting any other page or resource on the www.
We just click a link and the browser males a request and handles the
response. The content type (CSV) header tells the browser to treat the
response as a file, so it should show a download confirmation.

—
You are receiving this because you were assigned.
Reply to this email directly, view it on GitHub
<https://github.com/CiviWiki/OpenCiviWiki/pull/169#issuecomment-378838631>,
or mute the thread

delta-plus:
Tell me if I'm understanding this correctly. You want the export button to
be a link that points directly to the CSV data without any sort of
request/response action, right? I can do that using something like
render_to_response() and including a partial template into thread.html.
However, I'd still have to run some code when the thread loads to get the
right CSV data and if more civis are added or updated, I'd have to
automatically update the href again. That would cause more load on the
server than if the CSV was just given to the user if and when they wanted
it, so I don't think I'm getting exactly what you mean.

brylie:
For clarification, when the user clicks the button, the browser will make a GET request to the CSV endpoint. That is when the CSV will be generated and returned, with a header to tell the browser it is a CSV file. That is how the Django docs recommend [Outputting CSV with Django](https://docs.djangoproject.com/en/2.0/howto/outputting-csv/).

delta-plus:
Yeah, I just tried it out. It wasn't working earlier with the POST
request's response, so I had to do the whole invisible href data thing.
This is much better, thank you.

delta-plus:
Alright, it should be good to go now.

project/webapp/templates/thread.html
```diff
@@ -90,4 +90,12 @@
     //
     // if (cw.ws.readyState == WebSocket.OPEN) cw.ws.onopen();
 </script>
+
+<script>
+  $(document).ready(function() {
+    // Add current thread ID to export button href
+    document.getElementById('csv-export').href = '/civi2csv?thread=' + /[^/]*$/.exec(window.location.href)[0];
```
brylie: This is not necessary either. You can pass in the `threadId` as a URL parameter, defined in the route. See comment above.

project/frontend_views/urls.py
```diff
@@ -16,4 +16,5 @@
     url(r'^invite$', v.invite, name='invite'),
     url(r'^beta_register/(?P<email>[\w.%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,4})/(?P<token>\w{31})$', v.beta_register, name='beta_register'),
     url(r'^$', v.base_view, name='base'),
+    url(r'^civi2csv$', v.civi2csv, name='civi2csv'),
```
brylie: Change this route to pass in `threadId` in the URL like so:

```python
url(r'^thread/(?P<thread_id>\w+)/csv$', v.civi2csv, name='civi2csv'),
```

delta-plus:
Okay, will do.

delta-plus:
Alright, so everything works except the relative href. Setting href to
"csv" redirects to "/thread/csv" instread of "/thread/[thread ID]/csv".
This seems to be standard behavior. I could use a base tag, but that would
still require JavaScript to set the base href to the thread ID. Otherwise,
we'd need to make sure that threads are accessed with a trailing slash in
order for the relative URL to work as planned.

brylie:
The `id` should be part of the returned thread object that is being rendered by the template. You can use it to construct the `href` URL, using handlebars syntax:

```
<a 
    id="csv-export"
    href="thread/{{this.model.get('id')}/csv"
    class="btn waves-effect">
    Export
</a>
```

delta-plus:
Ah, I see. Alright, I'll try it out. Thanks.

delta-plus:
Hmm. It turns out {{this.model.get('id')}} doesn't work. I can get things
like the title and the author of the thread, but not the ID. I checked
api/models/thread.py and it looks like the ID isn't a direct property of
the thread model. Maybe I could get it through the thread manager somehow?

brylie:
In Django, `id` is part of every model, unless it is explicitly defined. So, the `id` property is probably being dropped when passing the thread from the backend to the client:

https://github.com/CiviWiki/OpenCiviWiki/blob/91887f07497590d69500bfa677147e9616ffe72a/project/api/read.py#L130-L146

Adding the following to that object should fix things:

```py
id: t.id
```

brylie:
> Do you mean I should add that to api/models/thread.py?

[Django models automatically get an `id` field](https://docs.djangoproject.com/en/2.0/topics/db/models/#automatic-primary-key-fields) (unless explicitly defined).

The [suggestion](https://github.com/CiviWiki/OpenCiviWiki/pull/169#issuecomment-379989903) was to pass the `id` property to the frontend via the REST endpoint, as it is currently being 'dropped'.

delta-plus:
I added the t.id to read.py, but {{this.model.get('id')}} still renders
nothing. I did try a couple other things, like getting the primary key with
.pk, but I didn't have any luck with those either.

brylie:
What does the `t` dict contain in the server code? Try [inspecting a thread document in the django shell](https://stackoverflow.com/a/6190203/1191545), to see whether it contains the `id` property`. This is just a matter of debugging and finding where/how to pass the thread ID to the client.

delta-plus:
I noticed that the issue_thread view is passing thread_id to the
thread.html template as part of a dict called data. It gets the thread_id
from the url the same way we have our URL set up. I think we'll either need
another small view to return thread_id to thread_body.html or one line of
JavaScript in thread.html which already has a script with "threadId =
'{{thread_id}}'" in it.

brylie:
OK, it seems like we should be able to pass the thread ID to the client as part of the response object, if it is not already being passed. Can you point to the lines in the files you mention?

delta-plus:
Sure. Relevant lines are:

Line 12 in frontend_views/urls.py
Lines 87, 133, and 139 in frontend_views/views.py
Line 62 in webapp/templates/thread.html

brylie:
> Line 12 in frontend_views/urls.py 
> Lines 87, 133, and 139 in frontend_views/views.py 
> Line 62 in webapp/templates/thread.html

Those are all on the frontend. See if you can make a change in the backend that will pass the value to the frontend. I really believe this is just a matter of sending a complete thread object from the backend.

brylie:
What do you see in your shell when you `print t` on the following line?

https://github.com/CiviWiki/OpenCiviWiki/blob/91887f07497590d69500bfa677147e9616ffe72a/project/api/read.py#L129

delta-plus:
I actually just got it. Had to add ' "thread_id": thread_id, ' to the
thread_wiki_data dict in the issue_thread view. I'll edit the PR in just a
sec.

delta-plus:
Also, I'd like your input on my last Riot message, please.

delta-plus:
Okay, just updated the pull request. Thanks again for reviewing! Let me
know if anything else needs to done with it.

delta-plus:
Okay, now it should be ready.

project/webapp/templates/thread.html
```diff
@@ -90,4 +90,5 @@
     //
     // if (cw.ws.readyState == WebSocket.OPEN) cw.ws.onopen();
 </script>
+
```
brylie: Go ahead and delete this line too. It should basically look like this file was untouched in the PR, since we no longer have any changes here.

project/frontend_views/views.py
```diff
@@ -266,3 +268,18 @@ def about_view(request):
 
 def support_us_view(request):
     return TemplateResponse(request, 'static_templates/support_us.html', {})
+
+@csrf_exempt
+def civi2csv(request, thread_id):
```
brylie: Add a [docstring](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring) to this function, describing its overall purpose.

brylie:
Really close! Just a couple more things to do for clean and well-documented code.

delta-plus:
Alright, just added the changes now.

PR #185, opened 2018-04-24
2to3

### LINKED ISSUE #18: Update project to use Python 3
CiviWiki seems to depend on Python 2.7. What are some of the reasons for staying with 2.7?

# Goal
Support the latest features of Python, such as `async` / `await`, by updating CiviWiki source/dependencies to support Python 3.

delta-plus:
Closes #18

This pull request updates all Civiwiki code to be Python 3 compliant. In order to maintain compatibility, there are a few rules to follow when writing Python code:

- Use print() instead of print.

- Use nice Python 3 Unicode strings, like so:
  from __future__ import unicode_literals
  from builtins import str

- Generally, just be aware of other Python 3-isms, like how 5/2 is 2.5 instead of 2, etc.

The files should be good to go, but a review is certainly a good idea. Thank you in advance, noble reviewers! If there is more info you think should be added here or somewhere else to help maintain Python 3 compliance or if something is incorrect, please drop a comment.


delta-plus:
So, just to be sure, do you want all the files in project/env removed or
just some files specifically?

jl24:
@delta-plus the entire project/env/ directory should be removed, as well as any .bak files that were generated.

delta-plus:
Okay, cool. Will do.

project/api/models/account.py
```diff
@@ -201,7 +207,7 @@ def resize_profile_image(self):
         Resizes and crops the user uploaded image and creates a thumbnail version of it
         """
         profile_image_field = self.profile_image
-        image_file = StringIO.StringIO(profile_image_field.read())
+        image_file = io.StringIO(profile_image_field.read())
```
jl24: using io.StringIO won't work since it is designed to handle Unicode string buffer. For this case, which is to handle data from an image file, we will have to use `io.BytesIO()`

project/api/forms.py
```diff
@@ -1,14 +1,17 @@
...
 from django.contrib.auth.models import User
-from models import Account
+from .models import Account
 
 class UpdatePassword(forms.ModelForm):
     """
     Form for updating User Password
     """
-    class Meta:
+    class Meta(object):
```
jl24: I believe django calls for the old-style class calls with just `Meta` instead of `Meta(object)`. Best to revert this particular change

project/api/models/civi.py
```diff
@@ -163,7 +169,7 @@ def score(self, request_acct_id=None):
         f = 0 #TODO: favorite val, is 'favorite' a meaningful name for this variable?
 
         # Calculate how long ago the post was created
-        time_ago = (current_time - post_time.replace(tzinfo=None)).total_seconds() / 300
+        time_ago = old_div((current_time - post_time.replace(tzinfo=None)).total_seconds(), 300)
```
jl24: I dont think we have any integer division we want to keep in the score function so you can revert all the `old_div` calls back to `/`

GokuMizuno:
The additions from Propublica #214 are all Python 3 compliant, so after that is approved to be merged, this branch can be merged as well, without any problems.

brylie:
Changes look good to me, so I approved PR. Before merging:

1. resolve conflicts in
    -  project/api/models/civi.py
    - project/api/urls.py
2. run QA on the code, full test suite and a bit of manual testing

brylie:
Also, when a PR closes an issue, make sure to add the following text to the PR description:

> Closes #18 

Substitute the actual issue number in other cases, but #18 in this PR description.

jl24:
@GokuMizuno Could you resolve the conflicts?

project/api/models/civi.py
```diff
@@ -105,8 +105,15 @@ class Civi(models.Model):
     votes_pos = models.IntegerField(default=0)
     votes_vpos = models.IntegerField(default=0)
 
-    def votes(self):
-        from .activity import Activity
+    def __str__(self):
+        return self.title
+
+    def __unicode__(self):
```
brylie: I believe with Python 3, all strings are unicode by default. So `__unicode__` is deprecated. Please follow the [directions outlined in Django documentation](https://docs.djangoproject.com/en/1.11/topics/python3/#str-and-unicode-methods).

project/api/models/civi.py
```diff
@@ -190,8 +197,10 @@ def score(self, request_acct_id=None):
             votes_total = votes['total'] if votes['total'] > 1 else 2
 
             #step3 - C
-            if abs(x) / v <= 5:
-                rank = abs(scores_sum) * math.log10(votes_total) * amp + y + f + g / time_ago
+
+            if abs(scores_sum) / votes_total <= 5.0:
+                rank = abs(scores_sum) * math.log10(votes_total) * amp + scores_sum + favorite + gravity / time_ago
```
brylie: While we're here, lets see if we can give `amp` a slightly more meaningful name. If not, it's alright, but this may be an easy readability improvement.

PR #194, opened 2018-05-18
Share Button

### LINKED ISSUE #150: Sharing Civi Threads
1) Ensure individual thread addresses are unique and sharable 
2) Add sharing button to the thread (viewable by all and including multiple sharing options -- Social media, via email, etc.) 
3) Allow users to view a thread without logging in. 
4) Add pop-up asking the user to sign up to save their voting data.

delta-plus:
Closes #150 Sharing Civi Threads. This pull request would add a button to the right of the Export button that contains a dropdown menu with links for sharing the current thread URL to social media and email. In links with an appropriate URL parameter, the thread title will also be automatically included in the shared content. If you'd like the buttons to be in a different order or if you think some should be added or removed, please add a comment.


project/webapp/templates/partials/thread/thread_body.html
```diff
@@ -41,6 +41,144 @@
                             <span class="text">Export</span>
                         </a>
                     </div>
+                    <div class="add-button" style="margin-left: 4px; position: relative">
```
brylie: See if you can encapsulate this into a `share_buttons` partial template, and include it here. That way it is reusable, and keeps the top level threads template relatively clean.

project/webapp/templates/partials/thread/thread_body.html
```diff
@@ -76,4 +214,7 @@
         </div>
     </div>
 </div>
+
+<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js">
```
brylie: Script imports should be in the document `head` or near the bottom of the top-level template. Also, be considerate when using *any* third-party code, particularly from Google and Facebook, as it may reduce people's privacy, or worse.

project/webapp/templates/thread.html
```diff
@@ -53,6 +53,16 @@
 <body>
     {% include "partials/utils/global_nav.html" %}
     <div id="thread"></div>
+
+    <div id="fb-root"></div>
+    <script>(function(d, s, id) {
+        var js, fjs = d.getElementsByTagName(s)[0];
+        if (d.getElementById(id)) return;
+        js = d.createElement(s); js.id = id;
+        js.src = 'https://connect.facebook.net/en_US/sdk.js#xfbml=1&version=v3.0';
```
brylie: As requested during our weekly meeting, please do not embed third party code in our templates, particularly from Facebook. Facebook uses these beacons to track and profile people. Using these beacons in our website makes us complicit with the violation of people's privacy.

project/webapp/templates/thread.html
```diff
@@ -90,4 +100,53 @@
...
+
+    $('#share-menu').click(function(event){
+        event.stopPropagation();
+    });
+
+    // Adds current thread URL and title to sharing link
+
+    links = document.getElementsByClassName('resp-sharing-button__link');
+    titleVars = ['text=', 'title=', 'caption=', 'subject='];
+    for (i = 0; i < links.length; i++) {
```
brylie: Ok, this code is clever, but there is hopefully a simpler solution. I.e we have two nested loops, index checking, array indexing, nested property access, string slicing and concatenation. While this seems like it would work, it also may prove to be brittle and hard to debug.

Is there a way we can construct the `share_url` on the backend, with simple concatenation. Then pass it in to the `share_buttons` template as a context variable (for client rendering).

brylie: If we can't find a 'cleaner', 'simpler' solution here, this one is alright. Instinct tells me there may be other approaches here.

delta-plus:
Okay. The Facebook script was something I forgot to remove from one of the
earlier button implementations. I'll get rid of it and see about a template.

brylie:
If the template approach doesn't work out, we can revert to this solution (without third party codes).

brylie:
Closing PR as work is ongoing in a different PR. (We can re-open if needed)

delta-plus:
Closing it is fine. The other PR should be done.

PR #196, opened 2018-05-22
Update current About Us Page

### LINKED ISSUE #117: Add Current Development Team to About Us Page
If you feel comfortable being advertised as part of the CiviWiki team, upload a picture and a title to the repository so we have the assets to update out about us page.

jl24:
Resolves #117


project/webapp/templates/static_templates/about.html
```diff
@@ -9,11 +9,11 @@
 </head>
 
 <script id="about-template" type="text/template">
-    <header>
+<header>
     <div class="content-header center">
         <div class="container section">
             <h2 class="title">About Us</h2>
-            <p class="description">The team that is bringing you CiviWiki</p>
+            <p class="description">The team that is bringing you civiwiki</p>
```
brylie: I believe 'civiwiki' should be camel-case, i.e. 'CiviWiki'. At least it is camel-case in other places, such as the mission statement. In any case, we should be consistent.

brylie: In effect, CiviWiki is the name of the organization and software, so should probably be camel-case.

jl24:
@brylie I've fixed the description errors

PR #197, opened 2018-05-22
Template-based thread sharing button

### LINKED ISSUE #150: Sharing Civi Threads

delta-plus:
Closes #150 Sharing Civi Threads. This pull request would add a button to the right of the Export button that contains a dropdown menu with links for sharing the current thread URL to social media and email. In links with an appropriate URL parameter, the thread title will also be automatically included in the shared content. This is the same code as on the share-button branch, just cleaned up and reformatted to use the template structure. This means no jQuery or DOM manipulation is used for adding the correct links now!


project/webapp/templates/partials/thread/thread_body.html
```diff
@@ -41,6 +42,12 @@
                             <span class="text">Export</span>
                         </a>
                     </div>
+<<<<<<< HEAD
```
brylie: Clean up these merge comments.

project/webapp/templates/thread.html
```diff
@@ -90,4 +94,36 @@
     //
     // if (cw.ws.readyState == WebSocket.OPEN) cw.ws.onopen();
 </script>
+<<<<<<< HEAD
```
brylie: Clean up merge comments.

project/webapp/templates/thread.html
```diff
@@ -90,4 +94,36 @@
...
+    function share() {
+        if (document.getElementById('share-options').className != 'share') {
+            document.getElementById('share-options').className = 'share';
+        } else {
+            document.getElementById('share-options').className = 'share-hidden';
+        }
+    }
+</script>
+
+>>>>>>> 9c32f602... Updated share button to use template format
```
brylie: Clean merge comments.

brylie:
The JS is a lot cleaner in this PR, much easier to grok! :+1:

delta-plus:
Thanks! It was good to learn about the template system you guys use.
Interesting stuff. I can definitely see the benefits. The code should be
good to go now.

project/webapp/templates/thread.html
```diff
@@ -90,4 +94,33 @@
     //
     // if (cw.ws.readyState == WebSocket.OPEN) cw.ws.onopen();
 </script>
+
```
jl24: All of this logic should be moved to the thread.js file. Use backbone.js events to attach and handle events. So instead of being in separate script tags, the code should be using the backbone view event structure (http://backbonejs.org/#View-events) .

jl24:
I'm getting huge unstyled black buttons. Seems like some styling code is missing? 

Also, I personally don't think we need the Tumblr, Google+ and WhatsApp buttons, but wondering what everyone thinks

delta-plus:
Oh, yeah. Sorry. I must not have pushed thread.less. I'll do that in a
couple hours, once I get home.

delta-plus:
Okay, added the styling. As for the Backbone.js events, I'll leave that for another contributor. Sorry I can't clean it up a bit more, I just have a lot going on right now.

jl24:
@delta-plus That's fine then. I'll pick it off where you left off.

delta-plus:
Cool. Thanks again for all the info and everything! I'll have to check out
the site later on and see how things progress. It's been a pleasure!

jl24:
The code has been restructured to conform to the backbone.js standards. I've removed Tumblr because the link was actually not functional, and WhatsApp because our current platform isn't exactly mobile ready.

project/webapp/static/less/thread.less
```diff
@@ -126,18 +128,17 @@ body {
...
-
-
                 margin-left: auto;
-                .btn{
+                
+                .btn {
                     font-size: 12px;
                     padding: 0 8px;
                     margin-top: 4px;
+                    margin-left: 4px;
```
brylie: Lets try and use relative units for layout styling. E.g. `em` or `rem` values rather than `px`.

PR #199, opened 2018-05-28
Civi Model score function undefined variables bugfix

### LINKED ISSUE #198: NameError: global name 'x' is not defined
Undefined variable in line 191 in `api/models/civi.py`. Prevents any pages with civi scoring from loading

### Solution
`x` should be `scores_sum`
`v` should be `votes_total`

jl24:
fixes #198

### DIFF
--- a/project/api/models/civi.py
+++ b/project/api/models/civi.py
@@ -155,16 +155,16 @@ class Civi(models.Model):
 
         if request_acct_id:
             account = Account.objects.get(id=request_acct_id)
-            y = (1 if self.author in account.following.all().values_list('id', flat=True) else 0)
+            scores_sum = (1 if self.author in account.following.all().values_list('id', flat=True) else 0)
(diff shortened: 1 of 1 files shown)


project/api/models/civi.py
```diff
@@ -188,7 +188,7 @@ def score(self, request_acct_id=None):
             votes_total = votes['total'] if votes['total'] > 1 else 2
 
             #step3 - C
-            if abs(scores_sum)/votes_total <= 5:
+            if abs(scores_sum) / votes_total <= 5:
                 rank = abs(scores_sum) * math.log10(votes_total) * amp + y + f + g / time_ago
```
brylie: As an aside, it might be useful to give meaningful names to `y`, `f`, and `g`.

jl24:
@brylie variable name has been fixed

PR #219, opened 2018-08-21
Update developer photos

jl24:
Add photos of Brylie and Gordon to our about us page.


jl24:
@brylie I've used your github profile photo for our main site. Is that alright? If not I will remove it.

PR #222, opened 2018-10-02
Closes #221

### LINKED ISSUE #221: Update pyca / cryptography  for security
The pyca / cryptography package has a security update. Please add the newest version ASAP.

dlarner3194:
Updates version number for pyca/cryptography package. Due to some security issues the package needed to be updated.

Package was updated to 2.3.1 

For reference: [cryptography](https://cryptography.io/en/latest/changelog/#v2-3-1 )

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,7 +20,7 @@ configparser==3.5.0
 constantly==15.1.0
 coreapi==2.3.3
 coverage==4.4.1
-cryptography==2.1.3
+cryptography==2.3.1
 daphne==1.3.0
 dj-database-url==0.4.2
 Django==1.11.7


dlarner3194:
There was one issue I ran into with updated the package and reinstalling the requirements text file...

I was getting a segmentation fault (core dumped) error when reinstalling. From some research it looks like its a Linux issue where pip was crashing on the install. Is that normal because of the amount/size of the packages?

brylie:
I'm not sure that it is normal for pip to crash when installing CiviWiki dependencies. Can you provide steps to reproduce?

In any case, is this PR ready for review?

PR #225, opened 2018-10-23
Making state/location visable from feed page

terrencetuy:
Adjusted the thread cards on the feed page to display state/location data. Copied the style that location information is displayed on the breadcrumbs when viewing at thread.

### DIFF
--- a/project/api/models/thread.py
+++ b/project/api/models/thread.py
@@ -28,7 +28,6 @@ class ThreadManager(models.Manager):
             "summary": thread.summary[:thread_truncate_length] + (ellipsis_if_too_long),
             "created": "{0} {1}, {2}".format(month_name[thread.created.month], thread.created.day, thread.created.year),
             "category_id": thread.category.id,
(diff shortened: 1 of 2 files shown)


project/api/models/thread.py
```diff
@@ -28,6 +28,8 @@ def summarize(self, thread):
             "summary": thread.summary[:thread_truncate_length] + (ellipsis_if_too_long),
             "created": "{0} {1}, {2}".format(month_name[thread.created.month], thread.created.day, thread.created.year),
             "category_id": thread.category.id,
+            "state": thread.state if thread.level == "state" else "federal",
```
brylie: Just a thought, but could `state` allow Null values?

terrencetuy: Sorry, the 'state' property isn't being used (it was a false start on my part). I'm using the location parameter instead. I've updated the code to remove the state paramter.

The location parameter does handle having a null state. It'll just print "state". Hope this helps. Thanks.

PR #262, opened 2019-01-17
Initial package upgrade debugging

brylie:
Add fixes after initial package updates.


brylie:
Oh man! PEP 8 has spoken!!

brylie:
I have worked through several errors. We may just need to deprecate Channels for the timebeing, as it only seems to be used for real-time threads.

cc: @mdwest @jl24

brylie:
Also, I thought we had fixed most/all of those lint errors in a previous PR. Hm...

PR #325, opened 2019-02-20
Add docker setup.

TomaszKolek:


TomaszKolek:
Will add readme shortly

TomaszKolek:
BTW -  I do not think 80 chars for a line is a good number. I'd set 120. 80 seems to be a bit historical (like command line editors)

brylie:
@TomaszKolek please resolve the conflicts in `requirements.txt`, so we can rebase this branch.

PR #329, opened 2019-02-24
fix aloe tests

Unit111:
@brylie it appears that during a refactoring the location of the chromedriver was lost. I added it in the project and is now working. In order to run the tests you should:
1. Install 'aloe'. It's already a part of the requirements.txt
2. Run the command 'aloe' inside the test_framework directory. Otherwise it will not find the feature files.
3. This PR adds the chromedriver executable again. If this location is not convenient for you you can move it to another place and edit the path in the __init__ file. Also, this instance is for Windows.
Google Chrome needs to be installed on the system you run the tests.
Some of the tests also fail but given it's being an year since I wrote them, this is to be expected.

### DIFF
--- a/project/test_framework/__init__.py
+++ b/project/test_framework/__init__.py
@@ -1,3 +1,4 @@
+import os
 from contextlib import contextmanager
 
 from aloe import around, world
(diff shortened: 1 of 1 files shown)


brylie:
I think @TomaszKolek removed the Chrome driver as part of his work to dockerize the project. @TomaszKolek could you review this PR, and see if these directions can be merged with your ongoing work?

brylie:
In any case, we should do this in a way that is platform agnostic, i.g. should not be Windows-centric. We mostly anticipate testing on Unix systems, like OS X and Ubuntu.

Unit111:
Unfortunate;y I don't know a way around this. You will need to download and use the proper chromedriver version depending on the OS you use:
https://chromedriver.storage.googleapis.com/index.html?path=73.0.3683.20/

TomaszKolek:
OMG, I didn't receive any notification about this PR.
Yes, I moved drivers to docker

brylie:
K, thanks. Can you please add instructions on how to execute the test suite from your docker image @TomaszKolek?

PR #332, opened 2019-02-26
Updated bill model

TomaszKolek:
I decided to make it this way to have full backward compatibility and still have normalized database

Also I grepped the code by legislation app and it seems its not being used, but if you know any usages let me know please.


project/api/models/bill.py
```diff
@@ -1,16 +1,24 @@
...
 
     title = models.CharField(max_length=1023)
     short_title = models.CharField(max_length=1023)
     short_summary = models.CharField(max_length=1023)
     number = models.IntegerField(default=0)
     b_type = models.CharField(max_length=63)
+    source = models.CharField(max_length=50, choices=BillSources.SOURCES, default=BillSources.SUNLIGHT)
 
-    # status
+    meta = JSONField(default=dict())
```
brylie: What will be stored in the Meta object? If possible, it would be nice to have explicit fields, to avoid logic to pick apart this object and check for attributes. Also, the structure of the object might change as the third party API changes. So, a generic field like `source_id` might be suffucient here.

TomaszKolek: Oh, thats good point that that thing might change. So let's move the logic to the method and store just urls and id that prolly won't change

TomaszKolek:
@brylie I provided the changes you requested.

I also found 2 migrations that weren't added. I added them

PR #333, opened 2019-02-26
Propublica API proxy view + CRUD Bills View

TomaszKolek:


TomaszKolek:
@brylie Its first part of api views - cleanup of the code

brylie:
OK, thanks. Can we fix some of these lint issues. Also, go ahead and bump up the maximum line length to 120, as you suggested earlier.

PR #335, opened 2019-02-26
Proxy api propublica

TomaszKolek:


brylie:
Hei, sorry for the late comment in this PR, but lets see if we can incorporate this library:

https://pypi.org/project/python-congress/

The reason being, it gives us a clean abstraction for ProPubluca APIs that we can use inside your proxy endpoints.

TomaszKolek:
I checked this library, but currently, this lib does not support search query. I'd go this implementation and think about replacement once this lib will be more suitable for us.

TomaszKolek:
About pep8speaks I created issue with it: https://github.com/OrkoHunter/pep8speaks/issues/105

But I think we can follow this issue separately since I fix most of pep8 issues here.

PR #342, opened 2019-03-04
Add possibility to add linked bills.

TomaszKolek:


project/api/migrations/0028_auto_20190304_1834.py
```diff
@@ -0,0 +1,25 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name='bill',
+            name='number',
+            field=models.CharField(max_length=20),
+        ),
+        migrations.AlterField(
+            model_name='bill',
+            name='source',
+            field=models.CharField(choices=[(b'sunlight', b'sunlight'), (b'propublica', b'propublica')], default=b'propublica', max_length=50),
```
brylie: We are deprecating the sunlight API. Unless there is data in the database from Sunlight, we no longer need it as an option.

project/webapp/static/js/thread.js
```diff
@@ -680,9 +680,11 @@ cw.NewCiviView = BB.View.extend({
         // $('.responses').height($('#new-civi-box').height() + $('.responses-box').height());
 
         this.magicSuggestView = new cw.LinkSelectView({$el: this.$('#magicsuggest'), civis: this.options.parentView.civis});
+        this.magicSuggestBillView = new cw.LinkSelectBillView({$el: this.$('#magicsuggestbills'), civis: this.options.parentView.civis});
```
brylie: Break this into multiple lines for readability.

brylie:
Minor changes requested. Otherwise, it looks quite good!

project/api/migrations/0028_auto_20190304_1834.py
TomaszKolek: Choices should be there in case we'd like to run admin panel functionality.
I'd keep it because debugging of potential problems for further developers might be really hard. I can comment appropriate line in the code about sunlight

project/webapp/static/js/thread.js
TomaszKolek: makes sense. sure.

PR #345, opened 2019-03-10
 Fix auth problem with registration

TomaszKolek:


brylie:
@TomaszKolek can you refactor the code to  fix the codeclimate issue? We can ignore Pyup until the dependency upgrade task.

TomaszKolek:
@brylie here we go

PR #348, opened 2019-03-12
Adding sentry

TomaszKolek:


project/civiwiki/settings.py
```diff
@@ -22,6 +25,11 @@ def get_env_variable(environment_variable, optional=False):
             raise ImproperlyConfigured(error)
 
 
+sentry_sdk.init(
+    dsn="https://4107ae45a36e495e8c4a85a4cbcdea00@sentry.io/1413260",
```
brylie: Should this be stored in an environment variable, rather than hard-coded in our sourcecode? Also, other CiviWiki deployments, if any, may not use this feature. So, it should be enabled only if the environment variable is present.

project/frontend_views/views.py
```diff
@@ -286,7 +286,7 @@ def civi2csv(request, thread_id):
     import csv
     thread = thread_id
     response = HttpResponse(content_type='text/csv')
-    response['Content-Disposition'] = 'attachment, filename=' + thread + '.csv'
+    response['Content-Disposition'] = 'attachment; filename=' + thread + '.csv'
```
brylie: While small, consider fixing this bug in a separate PR, with related issue. That makes it easier to track and understand why this code was added.

brylie: I.e. this PR should only focus on Sentry.

project/civiwiki/settings.py
TomaszKolek: Oh, you're right! That's a great catch, thank you!

project/frontend_views/views.py
TomaszKolek: Its in separate commit so in the history it'll be a different change.
Nevermind, if its easier to manage different PRs that's fine will do it.

PR #353, opened 2019-03-16
[BE] My bills tab

TomaszKolek:
I also fixed voting - it was buggy a bit

### DIFF
--- a/project/api/models/account.py
+++ b/project/api/models/account.py
@@ -18,6 +18,7 @@ from core.constants import US_STATES
 from .hashtag import Hashtag
 from .category import Category
 from .representative import Representative
+from ..serializers import BillSerializer
 
 # Image manipulation constants
 PROFILE_IMG_SIZE = (171, 171)
@@ -39,6 +40,7 @@ class AccountManager(models.Manager):
             "profile_image": account.profile_image_url,
             "followers": self.followers(account),
             "following": self.following(account),
+            "my_bills": account.get_voted_bills()
         }
         return data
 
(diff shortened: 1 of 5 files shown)


project/api/models/account.py
```diff
@@ -259,3 +261,28 @@ def is_full_account(self):
             return True
         else:
             return False
+
+    def get_voted_bills(self):
+        from .activity import Activity
```
TomaszKolek: its to avoid circular dependencies

project/api/write.py
```diff
@@ -153,52 +154,31 @@ def rateCivi(request):
...
-            voted_civi.votes_neg = voted_civi.votes_neg + 1
-            vote_val = 'vote_neg'
-        elif rating == "neutral":
-            voted_civi.votes_neutral = voted_civi.votes_neutral + 1
-            vote_val = 'vote_neutral'
-        elif rating == "pos":
-            voted_civi.votes_pos = voted_civi.votes_pos + 1
-            vote_val = 'vote_pos'
-        elif rating == "vpos":
-            # c.votes_vpos = c.votes_vpos + 1
```
TomaszKolek: it was buggy commented I think

project/api/write.py
```diff
@@ -153,52 +154,31 @@ def rateCivi(request):
...
+    }
 
-        activity_data = {
-            'account': account,
-            'thread': voted_civi.thread,
-            'civi': voted_civi,
-        }
+    activity_vote_key = 'votes_{}'.format(rating)
+    vote_val = 'vote_{}'.format(rating)
+    setattr(voted_civi, activity_vote_key, F(activity_vote_key) + 1)
```
TomaszKolek: It resolves a problem with voting race - the previous implementation could lose some data

project/api/models/account.py
brylie: Add that as a comment to the code, so we don't forget.

project/api/models/account.py
```diff
@@ -259,3 +261,28 @@ def is_full_account(self):
...
+
+        activities = Activity.objects.filter(account=self).prefetch_related(
+            'civi__linked_bills', 'civi__linked_civis__linked_bills'
+        )
+        supported_bills = []
+        opposed_bills = []
+
+        for activity in activities.iterator():
+            if activity.is_voted_negatively:
```
brylie: Can we rename this to `activity.is_negative_vote` / `activity.is_positive_vote`? It is a bit less verbose.

project/api/models/activity.py
```diff
@@ -36,3 +36,11 @@ class Activity(models.Model):
 
     created = models.DateTimeField(auto_now_add=True, blank=True, null=True)
     last_modified = models.DateTimeField(auto_now=True, blank=True, null=True)
+
+    @property
+    def is_voted_positively(self):
```
brylie: For an API, `is_positive_vote` and `is_negative_vote` sound a bit more 'natural'.

project/api/write.py
brylie: Add that as a comment to the code.

TomaszKolek:
@brylie review done

PR #357, opened 2019-03-23
Materialize.toast refractor

JonanOribe:
New function for the management of toast and its duration


brylie:
@JonanOribe thanks for this contribution. Which issue is this PR related to? Is it a bug fix?

JonanOribe:
Hi, is not a bug, just a little refractor that Code Climate has suggest for
a cleaner code. The same piece of code was at four different places, so I
just create a function that makes the same

PR #386, opened 2019-04-27
Gathering and displaying votes data

TomaszKolek:


project/api/models/representative.py
```diff
@@ -50,11 +48,27 @@ class Representative(models.Model):
...
+        if account:
+            votes = {}
+            bills = account.get_voted_bills(serialize=False)
+            supported_query = Vote.objects.filter(bill__in=bills["supported_bills"], representative=self)
+            opposed_query = Vote.objects.filter(bill__in=bills["opposed_bills"], representative=self)
+            votes["supported_the_same"] = supported_query.filter(vote="yes").count()
+            votes["supported_different"] = supported_query.filter(vote="no").count()
+
+            votes["opposed_the_same"] = opposed_query.filter(vote="no").count()
+            votes["opposed_different"] = opposed_query.filter(vote="yes").count()
```
brylie: Can we add `voted_same` and `voted_different` counts here? I think that is what the user is ultimately after, to see how much they have in common with a representative.

- voted same is sum of
    - supported the same
    - opposed the same
- voted different is sum of
    - supported different
    - opposed different

It is also a bit more natural of language to say "`Representative` *voted the same* as `User` this many times".

project/webapp/templates/partials/account/utils/representative_card.html
```diff
@@ -10,6 +10,12 @@
                 <div class="text-lato">Party: {{party}}</div>
             </div>
             <div class="subtitle-lato alignment">-</div>
+            <div class="section">
+                <div class="link-lato name"> Votes opposed like me: {{ votes.opposed_the_same }}</div>
```
brylie: I think these should be "boiled down" to "voted same" and "voted different" counts.

project/api/models/bill.py
```diff
@@ -27,6 +26,8 @@ class Bill(models.Model):
     created = models.DateTimeField(auto_now_add=True, blank=True, null=True)
     last_modified = models.DateTimeField(auto_now=True, blank=True, null=True)
 
+    is_voted_data_updated = models.BooleanField(default=False)
```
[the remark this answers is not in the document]
TomaszKolek: Hm... so we can store both - once we gather info about voting we do not need know when was the last update, just want to omit this bill during everyday script execution (commit2)

project/api/models/representative.py
TomaszKolek: yeah sure, makes sense - didn't know how to handle it -> your way is great!

project/webapp/templates/partials/account/utils/representative_card.html
TomaszKolek: 🆗

project/webapp/templates/partials/account/utils/representative_card.html
```diff
@@ -10,6 +10,10 @@
                 <div class="text-lato">Party: {{party}}</div>
             </div>
             <div class="subtitle-lato alignment">-</div>
+            <div class="section">
+                <div class="link-lato name"> Votes same: {{ votes.voted_same }}</div>
```
brylie: Woohoo! This is a lot easier to "grok", as a simple-minded human (me). 

Now, it would also be good to see this as a **percentage** (`votes.voted_same / (votes.voted_same + votes.voted_different)`), since it distills the idea to a single "similarity" number that is widely understood. 

Not saying it has to be part of this PR, but it would be helpful to the end user (and could serve as the basis of a match-making system where representatives are ordered by compatibility with the user -- a sort-of legislative Tinder).

project/api/models/representative.py
```diff
@@ -50,11 +48,36 @@ class Representative(models.Model):
...
+
+            voted_different_count = Vote.objects.filter(
+                bill__in=bills["supported_bills"], representative=self, vote="no"
+            ).count()
+            voted_different_count += Vote.objects.filter(
+                bill__in=bills["opposed_bills"], representative=self, vote="yes"
+            ).count()
+
+            votes["voted_same"] = voted_same_count
+            votes["voted_different"] = voted_different_count
```
brylie: Can we add a `votes["compatibility_score"] = voted_same_count / (voted_same_count + voted_different_count)`? I think this would be a good "summary statistic" to show the user.

TomaszKolek:
@brylie Current state is:
Screenshot 2019-04-28 at 12 30 16

Issue #514, opened 2019-05-21
Materialize Toast Message is shown too quickly to the user


richifuentes:
In the "Setup your location” when the message "We could not use your browser's location. Please try entering your location" is shown with materialize.toast, this message is only shown very briefly so the user cannot really understand what is going on. My suggestion is the time of materialize.toast should be increase or a permanent message should be shown to the user so that she/he knows what is happening.
![image](https://user-images.githubusercontent.com/9931749/58106711-d18fc480-7be8-11e9-960c-72c96bf6dbce.png)

brylie:
Great first issue! Would you like to extend the Materialize.toast timeout?

richifuentes:
Sure, would 10 seconds be ok?

JDRanpariya:
Is this issue solved?

Yogita98:
@brylie is this issue up for grabs?

brylie:
@Yogita98 yes, please open a pull request with your changes. :-)

PR #525, opened 2019-05-30
Materialize Toast Message is shown too quickly to the user #514

richifuentes:
Just have changed the timeout of Toast in some messages that otherwise the user cannot really see. Changed the timeout from 2 seconds to 10.


JOO33:
Moving away from JS. See #827

PR #721, opened 2020-07-27
Fix for #514, Materialize Toast Message is shown too quickly to the user

### LINKED ISSUE #514: Materialize Toast Message is shown too quickly to the user

Di-ref:
Closes #514

Extended the Materialize.toast timeout from 2 seconds to 9 seconds.


brylie:
@all-contributors please add @Di-ref for code

PR #769, opened 2020-10-01
Update docker-setup.md

gteti:
Fixed ip address for correct copy-past into the browser.


brylie:
@all-contributors please add @gteti for docs

PR #772, opened 2020-10-02
Add a default parameter to the get_env_variable function

BrianPainter:
Allow users to set defaults in the settings.py file for any items they are getting from an environment variable. Useful to have sane defaults if environment variables are not set.

### DIFF
--- a/project/civiwiki/settings.py
+++ b/project/civiwiki/settings.py
@@ -13,13 +13,15 @@ from django.core.exceptions import ImproperlyConfigured
 from sentry_sdk.integrations.django import DjangoIntegration
 
 
-def get_env_variable(environment_variable, optional=False):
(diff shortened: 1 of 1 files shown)


brylie (changes requested):
I'm just wondering whether we should be using `os.getenv()` instead of this bespoke function.

BrianPainter:
Are you opposed to an additional package: "django-environ"? It uses os.environ to read environment variables, and has two additional features: 1) includes same functionality as dj_database_url 2) allows user to optionally create .env file containing the environment variables they wish to use (easy to set environment variables for development).

brylie:
Django-environ sounds fine.

BrianPainter:
Included as part of #778

PR #776, opened 2020-10-03
Update manage.py

sampoprock:
The import should be global instead of in local scope.


BrianPainter:
this file is stock django. I'm not sure we would want to modify.

brylie (approved):

PR #873, opened 2021-05-09
Remove django environ

### LINKED ISSUE #864: Remove django-environ
We currently load environment variables using django-environ in settings.py. However, django-environment only supports Django up to around version 2. Since we aim to use the latest Django LTS (version 3.2 #848) we should remove django-environ.

Remove django-environ and use `os.getenv` instead. Any code using environment variables should be allowed to have a None value for the environment variable, with sensible defaults.

obafemiolorungbon:
fixes #864

### DIFF
--- a/project/civiwiki/settings.py
+++ b/project/civiwiki/settings.py
@@ -7,13 +7,13 @@ Production settings file to select proper environment variables.
 import os
 
 # False if not in os.environ
-DEBUG = os.getenv("DEBUG",False)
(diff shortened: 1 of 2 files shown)


brylie (approved):

brylie:
@all-contributors please add @obafemiolorungbon for code

brylie:
Thank you so much @obafemiolorungbon! :-)

Issue #876, opened 2021-05-14
Add i18n tags to all Django templates


brylie:
As we port templates from the SPA app over to regular Django templates, ensure that we add localization support for all static texts. Prior to the 2.0 release, review all templates to make sure the internationalization is complete.

mharyam:
Hello @brylie, 
Can I take this?

brylie:
@mharyam yes please :-) I'll assign you to the issue. Please let us know how we can be of assistance.

mharyam:
Thank you, I will.

mharyam:
Hello @brylie,

For this task, do you think it is ideal I add template tag {%trans%} to all the individual templates?

brylie:
@mharyam, yes. Follow the [guidelines from Django documentation](https://docs.djangoproject.com/en/3.1/topics/i18n/).

mharyam:
Hello @brylie 
I have created my first pull request for this task
Link: https://github.com/CiviWiki/OpenCiviWiki/pull/976

I added translation to the landing page, header and footer 
If it is fine, I will proceed to the next one 

Thank you.

brylie:
Thank you so much @mharyam! :smiley: 

I've reviewed your pull request and made some minor suggestions with code changes that you can automatically accept.

mharyam:
Hello @brylie 

Thanks a lot for taking out time to review! I have accepted all the suggestions. 

It looks like you still need to merge the pull request, so it can be updated to the main branch

Thank you.

mharyam:
Thank you @brylie  for taking out time to approve the pull request. 

Do you think there are other pages I need to add translation to?

brylie:
Basically, just look for a `templates` subfolder in any of the Django app folders. For example, there are several [accounts-related templates](https://github.com/CiviWiki/OpenCiviWiki/tree/develop/project/accounts/templates/accounts) that need i18n tags.

rsk2:
Hi @brylie, I'd like to help with this issue. Please let me know if there is any work that still needs to be done for this issue.

PR #883, opened 2021-05-18
Adds i18n tag to all templates.

abhiprojectz:
Adds i18n tag to all templates.

This fixes issue [#876](https://github.com/CiviWiki/OpenCiviWiki/issues/876)

### DIFF
--- a/project/webapp/templates/account.html
+++ b/project/webapp/templates/account.html
@@ -1,3 +1,4 @@
+{% load i18n %}
 <!doctype html>
 
 <head>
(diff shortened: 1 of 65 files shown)


brylie:
@abhiprojectz issue #876 is intended to remind us to use the `i18n` library in all template texts after we port the templates over to Django syntax in issue #889. Would you like to assist with issue #889?

brylie:
I converted this pull request to draft, since it is not yet time to internationalize the template texts. We first need to port the templates from Handlebars to the Django template syntax #889

gorkemarslan:
Since we put together html files we don’t have such file structure in the project anymore.

PR #894, opened 2021-05-25
Use django-taggit for keywords

### LINKED ISSUE #854: Use django-taggit for Civi keyword hashtags
We currently use a `Hashtag` model to categorize Civis. Switch to using django-taggit for keyword tagging. Use the word "keyword" instead of "hashtag" throughout the UI and data model.

loukikstudent:
Closes #854


brylie (changes requested):
Remember to make the related migrations by running

```
python manage.py makemigrations
```

Once you have created the migrations, commit them to Git and push them to this branch.

brylie:
@loukikstudent try running CiviWiki without Docker. It should work reasonably well with just Python virtual environment. Try on a fresh version of the develop branch, as we have simplified the dependencies substantially today. Then, apply your changes against a fresh branch, so you don't have to wrestle with Git too much.

PR #900, opened 2021-05-26
Using django-taggit for keywords #854

### LINKED ISSUE #854: Use django-taggit for Civi keyword hashtags

loukikstudent:
Closes #854 

I have migrated the changes also added it under "INSTALLED_APPS" as well as updated requirements.txt.

The issue is in api/accounts.py -> Accounts model the field ai_interests, i cannot add TaggleManager() because it through an error during migrations
`ValueError: You can't have two TaggableManagers with the same through model.`
Because of this i cannot completely remove the Hashtag model.


brylie (approved):

brylie:
@all-contributors please add @loukikstudent for code

brylie:
@loukikstudent great work! PR is merged :-)

PR #904, opened 2021-06-05
Swapped the default Django User model with a custom model. And implemented user registration and login.

### LINKED ISSUE #855: Move "Account" model to own Django app
Currently, all of our models are defined under the `api` app. We plan to re-organize the code structure to follow a more conventional Django project structure and deprecate the frontend SPA.

Move the account-related code/models into a new `accounts` app.

## Related code
https://github.com/CiviWiki/OpenCiviWiki/blob/f152ad5986ebfd25ff60ad5ca7c34563818f1e17/project/civiwiki/urls.py#L24

https://github.com/CiviWiki/OpenCiviWiki/blob/f152ad5986ebfd25ff60ad5ca7c34563818f1e17/project/civiwiki/urls.py#L30

### Create custom user model
We can take this opportunity to create a custom user model. The project currently uses the default Django user model, so our custom user model can just inherit from `django.contrib.auth.models.User`

https://github.com/CiviWiki/OpenCiviWiki/blob/283460fef79dc69dc526c7ef5c29e4ec82bb9874/project/authentication/authentication.py#L3

See the Django docs for [customizing authentication with a substitute user model](https://docs.djangoproject.com/en/3.2/topics/auth/customizing/#substituting-a-custom-user-model).'

Update the `Account` model to use the custom user model, and move the `Account` model to the `accounts` app:
https://github.com/CiviWiki/OpenCiviWiki/blob/f152ad5986ebfd25ff60ad5ca7c34563818f1e17/project/api/models/account.py#L108

Make sure to update our registration form to use the correct user model:
https://github.com/CiviWiki/OpenCiviWiki/blob/f152ad5986ebfd25ff60ad5ca7c34563818f1e17/project/authentication/forms.py#L40-L42

### Use Django class-based authentication views
For the most part, all authentication logic is already written for us. Remove the custom authentication views and replace them with the built-in, [class-based authentication views](https://docs.djangoproject.com/en/3.2/topics/auth/default/#module-django.contrib.auth.views).

### User related forms and templates
We need to make sure users can register and change their passwords, at the very least. So, we need to create templates to render the registration forms.

https://github.com/CiviWiki/OpenCiviWiki/blob/f152ad5986ebfd25ff60ad5ca7c34563818f1e17/project/authentication/forms.py#L20

Our current registration form code may be a bit bloated, such as the sections on password validation which can otherwise be provided by Django password validators.

Remove the following:
https://github.com/CiviWiki/OpenCiviWiki/blob/f152ad5986ebfd25ff60ad5ca7c34563818f1e17/project/authentication/forms.py#L35-L38
https://github.com/CiviWiki/OpenCiviWiki/blob/f152ad5986ebfd25ff60ad5ca7c34563818f1e17/project/authentication/forms.py#L77-L84

This may be redundant:
https://github.com/CiviWiki/OpenCiviWiki/blob/f152ad5986ebfd25ff60ad5ca7c34563818f1e17/project/authentication/forms.py#L44-L50

### Passwords
Use Django built-in [password validators](https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-AUTH_PASSWORD_VALIDATORS) instead of custom code:
https://github.com/CiviWiki/OpenCiviWiki/blob/f152ad5986ebfd25ff60ad5ca7c34563818f1e17/project/authentication/forms.py#L66-L75

### CSRF
We probably need to re-enable cookie-based CSRF tokens, or use whatever is Django default:

https://github.com/CiviWiki/OpenCiviWiki/blob/283460fef79dc69dc526c7ef5c29e4ec82bb9874/project/civiwiki/settings.py#L60-L62

Consider whether it is desirable to disable CSRF on REST API views (or whether we will still support REST authentication after this pull request):
https://github.com/CiviWiki/OpenCiviWiki/blob/ff74546f8fdc277c9bc5705f3df0c807c1130c62/project/api/authentication.py#L5

a7madnassar:
Closes #855 

… user registration and user login using class-based views.


brylie (approved):

brylie:
Thanks, @a7madnassar! This is a great pull request. Very clean and Pythonic code.

@all-contributors please add @a7madnassar for code

Issue #848, opened 2021-04-20
Upgrade project to Django 3.2 LTS


brylie:
The latest LTS release of Django is 3.2. Upgrade CiviWiki to Django 3.2 and follow settings conventions set out in the Django 3.2 starter project.

https://docs.djangoproject.com/en/3.2/releases/3.2/

brylie:
@North15, would you be interested in taking this issue?

EkanshBhatnagar:
Hey I would be interested in taking this issue .

EkanshBhatnagar:
Could you guide me to the details.

brylie:
@EkanshBhatnagar thank you for your interest. @North15 has already agreed to take this task, so we should give them the opportunity first.

Would you be interested in a different task from our [2.0 milestone](https://github.com/CiviWiki/OpenCiviWiki/milestone/5)?

North15:
@brylie I'll start working on this in the afternoon, I'll let you know if I run into any issues.

North15:
@brylie I ran `sudo docker-compose up` and I received this error at the end.
`ERROR: Service 'backend' failed to build: The command '/bin/sh -c apt-get update && apt-get install -y --no-install-recommends   git   gcc   vim   build-essential   unzip   xvfb   libxi6   libgconf-2-4   default-jdk   supervisor &&   apt-get clean && /' returned a non-zero code: 126
`

North15:
I was able to run `sudo docker-compose up` and to have it complete successfully with a couple of changes in the Dockerfile. I'm working on the upgrade now.

North15:
So, from what I can tell when I installed the latest version of Django, and a few errors were thrown.
`
norser@norser-OptiPlex-3010:~/Documents/GitHub/OpenCiviWiki$ python3 -m pip install -U Django
`
`
ERROR: daphne 1.4.2 has requirement asgiref~=1.1, but you'll have asgiref 3.3.4 which is incompatible.
ERROR: channels 1.1.8 has requirement asgiref~=1.1, but you'll have asgiref 3.3.4 which is incompatible.
ERROR: asgi-redis 1.4.3 has requirement asgiref~=1.1.2, but you'll have asgiref 3.3.4 which is incompatible.
`

There were a few other dependencies that needed to be upgraded: `daphne`, `channels`, `agis-redis`, and `asgiref`(this one gets updated when Django is updated). I updated these as well to the latest version, but when I do that there is a cycle that starts to occur.

[...]

PR #910, opened 2021-06-07
Upgraded Django to 3.2.4

North15:
This is in reference to issue #848.

Please let me know if any other changes are needed.


a7madnassar:
@North15 , I think these changes will break the upgrade. For example 'available_attrs' does not exist in django.utils.decorators in Django 3.2.4, so this will break the install, and that's why it was removed in the first place. Do you know why it is needed and what it breaks exactly? I was able to run the app fine with the previous changes. Also, why did you remove 'DFAULT_AUTO_FIELD' from settings? It is needed with the new Django version because we define our models without the 'id' filed.

brylie (approved):

PR #875, opened 2021-05-13
Remove account location data + representatives data

tberm:
Closes #856 
Closes #830

This aims to remove all fields for user accounts that are related to geographical location, as well as all functionality related to elected representatives.

There are a small number of minor changes I had to make to get the app to run with Python 3.

Note that the process for a user creating a profile is still a multi-stage flow, despite the fact that there is now only one stage of actual data input (since the location stage is gone). This feels a bit odd, but I have left it how it is to keep this PR simple and because it may be desirable to add other steps in future.

PR #934, opened 2021-06-15
Remove google map dependencies and other related problems:

shashankks0987:
- Remove beta redirection on login
- Remove beta blocker annotation

Occurences of google_map_api keys are removed
User was redirected to /beta on login which is fixed now
beta blocker has been removed at certain places


project/webapp/templates/user/settings.html
```diff
@@ -119,15 +119,13 @@
...
-    var googleMapsApiKey = "{{ google_map_api_key }}";
-
     location_data = {
         lng: {{lng}},
         lat: {{lat}}
     }
 
     var mapModel = new cw.Map({coordinates: location_data});
-    var mapView = new cw.MapView({model: mapModel, googleMapsApiKey: googleMapsApiKey});
+    var mapView = new cw.MapView({model: mapModel});
```
brylie: I believe we should also remove the `mapView` and `location_data` since we are moving away from storing user locations at all. This code may have been accidentally left over from PR #875

brylie (changes requested):

brylie:
@shashankks0987 small changes requested. Also, please resolve the conflicts so we can merge this PR.

brylie:
Hei @shashankks0987 just following up. This pull request is great overall. We just need to resolve the conflicts. Would it be easier to re-open the pull request against the latest changes in `develop` or do you think the conflicts are relatively easy to resolve?

shashankks0987:
@brylie Yes I'll do it asap, caught up in some work

brylie:
Thanks @shashankks0987. I hope you don't feel rushed. I just want to make sure we merge your contribution :-)

project/webapp/templates/user/settings.html
shashankks0987: @brylie I've made this change here but I see that they are used at a few other places in some js. I'm not really confident of removing them though. I guess we should start having a test suite so that these changes can be made more confidently

brylie (approved):

brylie:
@all-contributors please add @shashankks0987 for code

PR #951, opened 2021-07-12
#939: Removed parameter from call to Djando `distinct` function b/c parameters invoke the SQL cmd DISTINCT ON and only PostgreSQL supports that cmd

### LINKED ISSUE #939: sqlite3 doesn't seem to support some db operations we use in our app
When running the app with sqlite in development on the /settings page, I'm getting the following error when trying to make changes to a users first and last name.

It seems like sqlite might be incompatible with some db operations we use in our app.

image

cc: @brylie

jabrielcjones:
Closes #939

The bug happens when the settings page tries to read a user's account info from the database. It uses Django's `distinct` function. If the `distinct` function is used with parameters, then it will invoke the SQL command DISTINCT ON which is only supported by PostgreSQL. @deepspaceghost figured this out while we were pair programming.

We haven't gotten a chance to fully regression test the fix. The DISTINCT ON command might have been used for a reason, and without it something else might be broken. There are also a lot of other similar calls to Django's distinct function that could be causing similar errors.

### DIFF
--- a/project/api/read.py
+++ b/project/api/read.py
@@ -54,7 +54,8 @@ def get_profile(request, user):
         voted_solutions = Activity.objects.filter(
             account=a.id, civi__c_type="solution", activity_type__contains="pos"
         )
-        solution_threads = voted_solutions.distinct("thread__id").values_list(
+
+        solution_threads = voted_solutions.distinct().values_list(
             "thread__id", flat=True
         )
 
@@ -90,10 +91,16 @@ def get_profile(request, user):
                 result["follow_state"] = True
             else:
                 result["follow_state"] = False
+
+        print(result)
         return JsonResponse(result)
 
     except Account.DoesNotExist as e:
         return HttpResponseBadRequest(reason=str(e))
+
+    except Exception as e:
+        print('HEY! HEY!')
+        return(e)
 
 
 def get_feed(request):
--- a/project/api/read.py
+++ b/project/api/read.py
@@ -55,9 +55,11 @@ def get_profile(request, user):
             account=a.id, civi__c_type="solution", activity_type__contains="pos"
         )
 
-        solution_threads = voted_solutions.distinct().values_list(
-            "thread__id", flat=True
-        )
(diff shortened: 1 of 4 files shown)


project/api/read.py
```diff
@@ -54,7 +54,8 @@ def get_profile(request, user):
         voted_solutions = Activity.objects.filter(
             account=a.id, civi__c_type="solution", activity_type__contains="pos"
         )
-        solution_threads = voted_solutions.distinct("thread__id").values_list(
+
+        solution_threads = voted_solutions.distinct().values_list(
```
pscheid2: I think you are calling distinct before limiting the query to the `thread__id` column here, so the results might not be limited to distinct thread__ids.

Per this [SO](https://stackoverflow.com/questions/3852104/select-distinct-individual-columns-in-django), does something like `voted_solutions.values("thread__id").distinct()` work? It seems like that would appropriately remove duplicate thread__ids from the results.

jabrielcjones: Let me give it a try

jabrielcjones: It looks like it works. I added it to the PR and read up on the `values` method so I'm confident it serves the original purpose. The risk-averse side of me still wants to get it tested more thoroughly. Thanks for the feedback!

brylie (approved):

brylie:
This looks good. Should the same pattern be applied to the other places where `distinct` is used?

https://github.com/CiviWiki/OpenCiviWiki/search?q=distinct

jabrielcjones:
> This looks good. Should the same pattern be applied to the other places where `distinct` is used?
> 
> https://github.com/CiviWiki/OpenCiviWiki/search?q=distinct

Probably. Should we make the change in this PR or the next?

pscheid2:
> Probably. Should we make the change in this PR or the next?

Whatever is easiest!

brylie (approved):

brylie:
@all-contributors please add @jabrielcjones for code

PR #980, opened 2021-09-01
Issue #847 fixed

### LINKED ISSUE #847: Restore SessionAuthenticationMiddleware
We aim to move away from having a heavy JavaScript front-end, preferring instead to use Django templates (and sprinkles of JS where needed). This means we can use SessionAuthenticationMiddleware.

This will also require restoring the default authentication classes in `settings.py`

vishalda:
Closes #847 

Changed DEFAULT_AUTHENTICATION_CLASSES to SessionAuthentication in settings.py


brylie:
Are you able to log in after changing this setting?

brylie (approved):

brylie:
@all-contributors please add @vishalda for code

PR #999, opened 2021-09-08
Update dependencies as separated requirements files

### LINKED ISSUE #998: Seperate requirements.txt
Now, our [requirements.txt](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/requirements.txt) file includes all the dependencies that are used in development, testing, and production. It is a good practice to separate dependencies according to in which stages they are used. Eventually, a package used in the development won't be installed in the production stage.

As far as I know, there are two conventions here:

1. Create a folder named `requirements` and place `base.txt`, `dev.txt`, `prod.txt` etc. inside it. (e.g. [EvalAI](https://github.com/Cloud-CV/EvalAI/tree/master/requirements))
2. Create files without the parent requirements folder. (e.g. [weblate](https://github.com/WeblateOrg/weblate))

- [ ] Choose a way to place requirements files (as mentioned above)
- [ ] Separate requirements.txt file
- [ ] Update docs (e.g. `pip install -r requirements.txt` needs to be changed.)

gorkemarslan:
Fixes #998

We can use `pip install -r requirements/dev.txt` to install dependencies for the development. This command installs the dependencies not only from `requirements/dev.txt` but also `requirements/base.txt`.

Since the` test_framework` directory is not active and also at the moment we are not using `selenium` and `aloe` packages, I put these packages in the `test.txt`. Later, we will remove these unnecessary files and packages.


brylie (changes requested):
This looks great!

One thing that would be helpful would be to update our contributing guide with instructions for specifying the `dev` dependencies when [installing the Python environment](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/CONTRIBUTING.md#install-requirements).

gorkemarslan:
I just realized that `black` package is not present in both the old and the new requirements files. I added it. Also, I updated the contributing guide as your request.

brylie (approved):

gorkemarslan:
@brylie, could you add me as a contributor to the project. :innocent:

brylie:
@all-contributors please add @gorkemarslan for code

PR #1015, opened 2021-09-17
Update forms.py

### LINKED ISSUE #1016: User object not retrieved.
### Description

When I try running the project locally and I try registering a user, It shows an error.

I expect the registration to work successfully.

_No response_

### Further details

image
image

gorkemarslan:
Fixes #1016 

A further update is needed after my last PR.

### DIFF
--- a/project/accounts/forms.py
+++ b/project/accounts/forms.py
@@ -62,7 +62,7 @@ class UserRegistrationForm(ModelForm):
 
         email = self.cleaned_data.get("email")
 
-        if User.objects.filter(email=email).exists():
+        if get_user_model().objects.filter(email=email).exists():
             raise forms.ValidationError(self.error_message["email_exists"])
 
         return email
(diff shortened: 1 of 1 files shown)


brylie (approved):

brylie:
Although it might not be relevant in this particular case, remember to use "Closes #<issue_number>" or "Fixes #<issue_number>" in order to link pull requests to the issues they close.

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

> You can link a pull request to an issue by using a supported keyword in the pull request's description or in a commit message (please note that the pull request must be on the default branch).
>
> - close
> - closes
> - closed
> - fix
> - fixes
> - fixed
> - resolve
> - resolves
> - resolved

AbhijithGanesh (changes requested):
Can you please use the model in the meta class, is there any downside to it?

gorkemarslan:
`get_user_model()` method gives you which user model is used in the project. It returns `AUTH_USER_MODEL` from the `settings.py`. In our case, it is defined such that

https://github.com/CiviWiki/OpenCiviWiki/blob/da0ad36db74b992efc7acc31823eb05bffe5582f/project/core/settings.py#L172

If `AUTH_USER_MODEL` were not specified here, the default Django `User` model is used for `get_user_model()`.

We shouldn't import the user model directly because there might be circular dependency issues, especially for ForeingKey and Many-to-Many fields usages.

This way is also recommended by Django official [docs](https://docs.djangoproject.com/en/3.2/topics/auth/customizing/#django.contrib.auth.get_user_model).

AbhijithGanesh:
Points well noted, thanks for clarification,

PR #1021, opened 2021-09-18
Merging the first 29 migrations, Optimized

AbhijithGanesh:
I have initially optimized files for API app otherwise the Accounts up will throw a circular import error.


AbhijithGanesh:
I believe, this PR can wait after my discovery of the newer issues, please donot merge this until we have tested the migrations fully, I don't want this to damage the repository.

AbhijithGanesh:
This PR was irrelevant and the next PR has all the relevant files

PR #1023, opened 2021-09-20
Checking out squashed migrations for 29 files in the api app

### LINKED ISSUE #1022: Bad Migration Tree
### Description

When I was trying to squash migrations of the api , there seems to be an issue related to the whole ancestry of migrations, I am working on this as a part of  #1017. The parent tree appears to be slightly displaced which I'll be looking forward to patch with the other issue mentioned.

_No response_

### Further details

image

AbhijithGanesh:
Closes #1022

Addresses issue #1022 without any damages to other migration.


AbhijithGanesh:
Once the commits are fine , we can remove other migrations related to the same.

AbhijithGanesh:
I would like the code of this PR to be tested before we merge this code. In my local code base, there isn't an issue migrating. Can @gorkemarslan and @brylie please let me know?

brylie:
Good idea.

@gorkemarslan are you available to check out this branch as well?

gorkemarslan:
@brylie, yes I am available for this purpose.

AbhijithGanesh:
It is not something big, in your test environment just run migrations. If it is fine (basically doesn't use api 0001-0029 ) it should be mergeable. We can even step ahead and remove the files (but In my honest opinion, I will let them stay just in case to jump back to old migrations)

gorkemarslan:
Apparently, everything's fine. Thank you for your effort @AbhijithGanesh.  :)

@brylie, you can approve merging this PR.

brylie (approved):

Issue #1032, opened 2021-09-23
Move all thread code from `api` to `threads` app


brylie:
The `api` app contains code that should be moved to the `threads` app. Move all of the following code to the `threads` app, fix any related imports, and make migrations (if necessary).

- [x] related code in `api/models/` to `threads/models.py`
    - [x] [Activity](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/project/api/models/activity.py)
    - [x] [Civi](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/project/api/models/civi.py)
    - [x] [Fact](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/project/api/models/fact.py)
    - [x] [Rationale](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/project/api/models/rationale.py)
    - [x] [Rebuttal](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/project/api/models/rebuttal.py)
    - [x] [Response](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/project/api/models/response.py)
    - [x] [Thread](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/project/api/models/thread.py)
    - [x] ThreadManager
    - [x] CiviManager 
    - [x] CiviImageManager
    - [x] CiviImage
    - [x] ActivityManager
- [x] related code in `api/read.py` -> `threads/api.py`
    - [x] [`get_thread`](https://github.com/CiviWiki/OpenCiviWiki/blob/db319670f20d2e266db5814ed03f9209105365ef/project/api/read.py#L125)
    - [x] [get_civi](https://github.com/CiviWiki/OpenCiviWiki/blob/db319670f20d2e266db5814ed03f9209105365ef/project/api/read.py#L189)
    - [x] [get_civis](https://github.com/CiviWiki/OpenCiviWiki/blob/db319670f20d2e266db5814ed03f9209105365ef/project/api/read.py#L201)
    - [x] [get_responses](https://github.com/CiviWiki/OpenCiviWiki/blob/db319670f20d2e266db5814ed03f9209105365ef/project/api/read.py#L213)
- [x] related code in `api/serializers.py` -> `threads/serializers.py`
    - [x] [CiviImageSerializer](https://github.com/CiviWiki/OpenCiviWiki/blob/db319670f20d2e266db5814ed03f9209105365ef/project/api/serializers.py#L120)
[...]

PR #1035, opened 2021-09-23
CodeQL Workflow

AbhijithGanesh:
Addition of yml files for CodeQL workflow checking.

### DIFF
--- a/project/threads/models.py
+++ b/project/threads/models.py
@@ -20,6 +20,7 @@ class Fact(models.Model):
     created = models.DateTimeField(auto_now_add=True, blank=True, null=True)
     last_modified = models.DateTimeField(auto_now=True, blank=True, null=True)
 
+
 class Category(models.Model):
     name = models.CharField(max_length=63)
 
@@ -30,7 +31,7 @@ class Category(models.Model):
         return self.name
 
     def __unicode__(self):
-        return self.name    
+        return self.name
 
 
 class ThreadManager(models.Manager):
@@ -391,8 +392,6 @@ class Civi(models.Model):
         return data
 
 
-
-
 class Notification(models.Model):
     account = models.ForeignKey(
         Profile, default=None, null=True, on_delete=models.PROTECT
@@ -450,6 +449,7 @@ class Response(models.Model):
 
     created = models.DateTimeField(auto_now_add=True, blank=True, null=True)
     last_modified = models.DateTimeField(auto_now=True, blank=True, null=True)
+
 
 image_upload_path = PathAndRename("")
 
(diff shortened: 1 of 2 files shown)


gorkemarslan:
@AbhijithGanesh, you change a `models.py` file so could you add the migrations that will come with the command of `python manage.py makemigrations`?

AbhijithGanesh:
Yes I can but that shall repeat for other parts too and I shall branch it in a better way. Here codeql and apideprecation got mixed up. I shall fix the same asap.

gorkemarslan:
I think we should not approve code changes in `models.py` files if related migrations are not added. 
But I would say you might need to remove api-deprecation commits from this PR and we approve workflow commits only.
For api-deprecation commits, you can attach migrations to your code changes in a new PR. (Or in the one you just closed.)
We can wait for @brylie, they better know it.

project/threads/models.py
```diff
@@ -1,3 +1,543 @@
...
+
+
+class Fact(models.Model):
+    body = models.CharField(max_length=511)
+
+    created = models.DateTimeField(auto_now_add=True, blank=True, null=True)
+    last_modified = models.DateTimeField(auto_now=True, blank=True, null=True)
+
+
+class Category(models.Model):
```
gorkemarslan, line 24: In issue #1032,  @brylie mentioned that 

> category - although this might be better in a categories app since it is shared between threads and accounts apps

You may not add `Category` model to `threads/models.py`.

brylie:
This pull request lacks context. What is CodeQL and how do we use it? At the very least we should have an open issue to discuss the benefits of this tool and possible alternatives. We may also need to update our project documentation with a note about the tool and its usage.

project/threads/models.py
```diff
@@ -1,3 +1,543 @@
+import os
```
brylie, line 1: As @gorkemarslan mentions, these changes should be in a separate pull request, since they are not related to the CodeQL workflow.

brylie (changes requested):

brylie:
Closing this pull request pending further discussion via a [Feature Request](https://github.com/CiviWiki/OpenCiviWiki/issues/new?assignees=&labels=community%2Ctriage%2Chelp+wanted&template=feature.yml&title=%3CAdd+a+brief+title+for+your+idea%3E+) or [Discussion topic](https://github.com/CiviWiki/OpenCiviWiki/discussions/new).

brylie:
For what it's worth, the CodeQL seems like a good addition. I would just like to explore the enhancement a bit further via discussion :smiley:

PR #1044, opened 2021-09-26
Fixed image upload

### LINKED ISSUE #392: When uploading a profile image failed then user needs to go back and click on Edit Profile again to upload new image
When users uploading a profile picture and this failed with the error message "Please use an image that 1280 x 960 pixels or smaller" then users need to go back to the profile and click on Edit Profile again in order to upload a new picture. Just clicking on Choose Picture doesn't do it.

wassafshahzad:
Fixed a few bugs in image uploads which were cause due to upgrading from python 2 to 3.

This PR Closes #392


gorkemarslan:
I think you mislabeled the issue as this PR (#1044). It should be issue #392. Right?

wassafshahzad:
> I think you mislabeled the issue as this PR (#1044). It should be issue #392. Right?

Yeah my bad,
I had the issue number in my PR name and still managed to mess it up
Thank you for the notification

project/accounts/models.py
```diff
@@ -169,10 +169,10 @@ def resize_profile_image(self):
         """
         Resizes and crops the user uploaded image and creates a thumbnail version of it
         """
-        profile_image_field = self.profile_image
-        image_file = io.StringIO(profile_image_field.read())
-        profile_image = Image.open(image_file)
-        profile_image.load()
+        #profile_image_field = self.profile_image
```
gorkemarslan: It is not a good practice to store unused code lines as commented. You can remove them.

brylie: As suggested, remove unused code.


```suggestion
```

project/accounts/models.py
```diff
@@ -169,10 +169,10 @@ def resize_profile_image(self):
...
+        #image_file = io.StringIO(profile_image_field.read())
+        profile_image = Image.open(self.profile_image)
+        #profile_image.load()
```
brylie: Remove unused code.
```suggestion
        profile_image = Image.open(self.profile_image)
```

brylie (changes requested):

project/accounts/models.py
wassafshahzad: Sure I will have this done by today

brylie: To make it easy, you can just press the "commit suggestion" button and the change will automatically be applied to the branch.

wassafshahzad: updated PR

brylie (approved):

PR #1045, opened 2021-09-26
Remove username and email labels (Fix #231)

### LINKED ISSUE #231: User Name and Email Forms Are Visually Bugged
During account creation, the username and email forms are visually bugged. 

The actual username and email is overlapped by the hint text (i.e. “User Name” and “Email) 

I expect these forms to be pre-filled with the User Name and Email the user selected on the previous screen and the hint text to only appear if they delete the text. 

image

gorkemarslan:
Fix #231 

Before:
Screen Shot 2021-09-26 at 19 23 06

After:
Screen Shot 2021-09-26 at 19 22 29

The problem is caused by overlapping labels and read-only fields. Removing labels fixes the problem. The problem is not browser-specific so it works for all browsers.

Note: In order to see changes on the browser, `python manage.py collectstatic` should be executed.


project/threads/templates/threads/partials/user-setup/step1.html
```diff
@@ -55,11 +55,9 @@
                             <div class="row">
                                 <div class="white-background-input-field input-field col s6">
                                     <input id="username" type="text" class="validate" value="{{username}}" disabled>
-                                    <label for="username">Username</label>
                                 </div>
```
wassafshahzad, line 59: @gorkemarslan . I think rather then removing labels all together we could add the class `active` on them to resolve the issue.
I think It would be better for consistency and accessibility.

gorkemarslan:
@brylie, I've just realized there are also problems in login and register forms. (as pointed out in [here](https://github.com/CiviWiki/OpenCiviWiki/issues/231#issuecomment-893174906)) The problem is caused by overlapping autocomplete by browser and placeholders. As far as I see (but not sure), this mostly occurs using materilize.css. I find a couple of solutions on Stack Overflow. I try to resolve this problem by updating CSS data (`login_less`) with adding webkit tool, solving with jQuery, adding `placeholder=" "`. Unfortunately, each solution and their combinations fail even though I spent some time. 

Maybe we can turn off `autocomplete` feature in HTML to delay the problem. Then you can approve the PR and close the issue. Or we can add subtasks to the original issue and mark resolved the first one. Another option, I can create a new issue. What is your suggestion?

project/threads/templates/threads/partials/user-setup/step1.html
gorkemarslan: It would be better, you're right. With my solution, labels disappear. In your solution, users can understand which is which for username and email fields. This would be much clearer. Thanks!

brylie:
> Maybe we can turn off autocomplete feature in HTML to delay the problem.

See the MDN article for a possible solution: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-autocomplete

gorkemarslan:
@brylie, actually the problem is only experienced by Chrome users. No matter what I try to do, the result is the same.

Mozilla:
Screen Shot 2021-09-27 at 17 33 54


Safari:
Screen Shot 2021-09-27 at 17 37 46


Chrome:
Screen Shot 2021-09-27 at 17 34 14


I try to use `autocomplete="off"` but it does not work. Removing autocomplete is not easy (possible) in Chrome as far as I understand. You can see [this SO question](https://stackoverflow.com/questions/12374442/chrome-ignores-autocomplete-off). Chrome constantly blocks `removing autocomplete` by new versions. I'm paraphrasing an answer:

>  In Chrome's perspective, nothing is more important than users and their preferences regardless of the intention of the website. Chrome always enhances user experiences.

So, I find spending more effort on this issue tiring.

gorkemarslan:
As @wassafshahzad suggested, the final appearance of the `setup` page:

Screen Shot 2021-09-27 at 12 30 02

wassafshahzad (approved):

wassafshahzad:
> As @wassafshahzad, the final appearance of the `setup` page:
> 
> Screen Shot 2021-09-27 at 12 30 02
> 

@brylie could you do a quick review aswell?

project/threads/templates/threads/partials/user-setup/step1.html
brylie: Yes. We need to keep the labels for the accessibility of the form.

brylie:
I wonder if this issue has been fixed upstream in Materialize. What happens if we update Materialize to 1.0.0?

Also, let's remove the Materialize CSS from our project and include it via CDN:

```html
<!-- Compiled and minified CSS -->
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/css/materialize.min.css">

<!-- Compiled and minified JavaScript -->
<script src="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/js/materialize.min.js"></script>
```

brylie:
> could you do a quick review aswell?

The profile page looks great!

brylie (approved):

Issue #1079, opened 2021-10-02
Restore and improve fixtures in `project/data/`


AbhijithGanesh:
Our project contains some [fixture data](https://docs.djangoproject.com/en/3.2/howto/initial-data/#providing-data-with-fixtures) that is useful to quickly populate our Categories and Threads models. However, we don't tell developers how to use the fixtures or that they even exist.

## Task

- [x] determine how to use fixtures to populate our initial data
- [x] test our existing fixtures to make sure they still work
  - [x] if needed, improve the fixture data so it works with any model changes 
- [x] #1081

## Original issue

<details>

<summary>original issue opened by @AbhijithGanesh</summary>

### Idea summary

Using one standard form of data 

### Further details

Currently we use both constants.py and .json to store common data rather than using both since it leads to confusion
```js
[{
    "fields": {
        "name": "Agriculture"
    },
    "model": "api.category",
    "pk": 0
},{
    "fields": {
        "name": "Environment"
    },
    "model": "api.category",
    "pk": 1
}, {
    "fields": {
        "name": "Energy"
    },
    "model": "api.category",
    "pk": 2
}, {
 ...
{
    "fields": {
        "name": "Other"
    },
    "model": "api.category",
    "pk": 21
}]

``` 
and you can view constants here
[Constants](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/project/core/constants.py)

</details>

AbhijithGanesh:
@brylie do we use the data from data directory anymore?

brylie:
The constants are used in Python code, so should remain as .py. Whereas the `.json` file is used to populate the Categories table with default categories, in the example above.

In both cases, I believe we are following established conventions.

Related documentation:

- Django docs - Providing initial data for models: [Providing data with fixtures](https://docs.djangoproject.com/en/3.2/howto/initial-data/#providing-data-with-fixtures)

brylie:
> do we use the data from data directory anymore?

I believe we can use this issue to improve our documentation. I.e. we can add a sub-section to our development guide describing how to use the `categories.json` and `sample_threads.json`.

AbhijithGanesh:
Got it, then I think the .json file should not reference api right?
We need to reference them to newer models right? @brylie

AbhijithGanesh:
can I work on this?

brylie:
> Got it, then I think the .json file should not reference api right?

@AbhijithGanesh, yes good catch! 

[...]

PR #1082, opened 2021-10-02
Fixing fixtures 

AbhijithGanesh:
This patch fixes parts of issue #1079 and allows load data method now

A small note from my end though:
- We've lost author_id and category_id in the [sample_threads.json](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/project/data/sample_threads.json) because of migrations that has occurred over time. We don't know who are the authors who initially came up with the ideas
- Those two fields have been deleted since we use  newer models.


AbhijithGanesh:
@brylie and @gorkemarslan can you please review this?

LightFelicis:
django.db.utils.IntegrityError: Problem installing fixtures: The row in table 'threads_thread' with primary key '1' has an invalid foreign key: threads_thread.author_id contains a value '2' that does not have a corresponding value in accounts_profile.id.

Shouldn't there be some placeholder authors added? :)

AbhijithGanesh:
Once my changes are merged, this issue shouldn't appear, check the files changed.

gorkemarslan:
@AbhijithGanesh, it is also possible to generate fixtures without providing `create` time-stamp field. The Django itself handles these timestamps when saving instances from fixtures. Unfortunately, StackOverflow is under maintenance right now, I am not able to send you useful links. This is not a change request now, just to inform you. :)

AbhijithGanesh:
I totally agree but the only proof which we can include(proof related to the person who created this) is timestamp, if we remove it, there won't be any trace of the previous author who contributed which is extremely unfair to those who contributed. @brylie

gorkemarslan:
You're right, nice catch. :)

AbhijithGanesh:
Please review

gorkemarslan:
@AbhijithGanesh, why did you remove `"category_id"` and `"author_id"` fields?

AbhijithGanesh:
I removed them because they're references to Primary key and their foreign key relation has been removed @gorkemarslan

project/data/sample_threads.json
```diff
@@ -1,201 +1,162 @@
...
-            "category_id": "1",
```
gorkemarslan, line 6: Could you restore `category_id` and `author_id` fields? Both category and author fields have `null=True` in their relations but it is more appropriate for a `thread` to have a `category` and an `author`.

gorkemarslan (changes requested):

project/data/sample_threads.json
gorkemarslan: @AbhijithGanesh, sorry for my late response. I was cloning your repo and look at your project. A thread should have these fields.

AbhijithGanesh:
Why so ? If they're removed what would it result on? @gorkemarslan

project/data/sample_threads.json
gorkemarslan: Could we continue the conversation here?

What do you mean by `removed`? (Did I miss a point? 🤔 )

Thread model has these fields.

```python
class Thread(models.Model):
    author = models.ForeignKey(
        Profile, default=None, null=True, on_delete=models.PROTECT
    )
    category = models.ForeignKey(
        Category, default=None, null=True, on_delete=models.PROTECT
    )
```

Let's change our perspective. Suppose one decides to create a new thread. What will this person do when creating a new thread? This person will write a title, a summary and choose a related category. In the end, the new thread will have all these fields. What we do with `fixtures` has to be the same. We just automize instance creation steps.

AbhijithGanesh: I agree with the part which relates to "It has to be the same ". But threads and categories donot have a categories_id right? How can I implement it if they don't exist? Should I query it first using dbshell and figure a way around or add a new column?

AbhijithGanesh: @gorkemarslan can you please help me with  this? I am a bit unsure about what to do

gorkemarslan: @AbhijithGanesh, I'm typing. I don't just follow your speed. 😅  Sorry my bad.

AbhijithGanesh: 😂😂😂 I am sorry but I was way too unsure about  to what to do and I really love this project

gorkemarslan: Whenever you create a new model instance. The instance has `id` (or `pk` field), which is an automatic increase integer number. Whenever you create a model instance on the database,  a new id is automatically assigned. The `id` field is a hidden field but can be overridden. As far as databases are concerned, id fields are a must. Think of your ID number. You're a person with an ID number for your government (in this case the database).  We call you with your name but you're registered in the system with your id. The ID number is meaningless for your friends (in this case the users of the website). On the other hand, officers (in this case backend developers) are interested in your ID number. 

You can look the [docs](https://docs.djangoproject.com/en/3.2/topics/db/models/#automatic-primary-key-fields), as well.

gorkemarslan:
```python
class Category(models.Model):
    name = models.CharField(max_length=63)

```

So a category model instance has four fields. `id`, `name`, `created`, and `last_modified`.

gorkemarslan: For example,
in **categories.json** file (in my branch it is not updated):
```json
{
    "fields": {
        "name": "Environment"
    },
    "model": "api.category",
    "pk": 1
}
```
**sample_threads.json:**
```json
{
        "model": "api.thread",
        "fields": {
            "title": "Agriculture Subsidization ",
            "summary": "...",
            "category_id": "1",
            "author_id": "2",
            "created": "2016-10-09 20:25:20.365392+00"
        }
    }
```
So this thread has a category with an id 1. (`"category_id": "1"`). You can see `categories.json` file, `categories_id: 1` means `Environment` category.

AbhijithGanesh: I understood where I went wrong, instead of removing author id I removed both the ids

AbhijithGanesh: @gorkemarslan I have patched my mistake, thanks for the amazing catch😄😅

AbhijithGanesh: Can you approve this @gorkemarslan

gorkemarslan: @AbhijithGanesh, `author_id` field should also be included. Could you restore it?

AbhijithGanesh:
But we don't know who the authors are @gorkemarslan , how will we map them?

AbhijithGanesh:
But we don't know who the authors are @gorkemarslan , how will we map them?

project/data/sample_threads.json
AbhijithGanesh: How will we map the author id to author? We don't know who wrote them right?

AbhijithGanesh: Is there a way to map the authors to their ids Gorkem Arslan? @gorkemarslan

AbhijithGanesh:
@brylie I think even this can be merged, I'm waiting for reviews

brylie:
@AbhijithGanesh, did you resolve the changes that @gorkemarslan requested?

AbhijithGanesh:
He suggested two, I resolved one and the other is not resolvable

project/data/sample_threads.json
gorkemarslan: @AbhijithGanesh, authors have already mapped Profile model via `author_id` field. 

```python
author = models.ForeignKey(
        Profile, default=None, null=True, on_delete=models.PROTECT
    )
```

So you should make any changes about `id` fields in the fixture data.

AbhijithGanesh: > Is there a way to map the authors to their ids Gorkem Arslan? @gorkemarslan 
> 
> 
They are mapped but they don't actually exist , if they don't exist, we can't load them

brylie:
> He suggested two, I resolved one and the other is not resolvable

I believe it is possible to resolve the other issue. As per the Django documentation:

> The most straightforward way of creating a fixture if you’ve already got some data is to use the `manage.py dumpdata` command
>
> https://docs.djangoproject.com/en/3.2/howto/initial-data/#providing-data-with-fixtures

So, the idea is to create the initial data directly in CiviWiki and then use the `dumpdata` command to generate fixtures with the appropriate foreign-key IDs.

brylie (changes requested):
I believe there is a solution for generating fixtures with correct foreign-key IDs.

As per the Django documentation:

> The most straightforward way of creating a fixture if you’ve already got some data is to use the manage.py dumpdata command
>
> https://docs.djangoproject.com/en/3.2/howto/initial-data/#providing-data-with-fixtures

So, the idea is to create the initial data directly in CiviWiki and then use the `dumpdata` command to generate fixtures with the appropriate foreign-key IDs.

https://docs.djangoproject.com/en/3.2/ref/django-admin/#django-admin-dumpdata

brylie:
@AbhijithGanesh I have requested changes with a suggestion for how to make the fixtures with the correct foreign-key relationships.

AbhijithGanesh:
@brylie and @gorkemarslan I have made new fixtures yet it is unable to solve the issue with author_id, it is because of this:

- author_id is an actual reference to the parameter of author not a mapping which will be implemented
- if AUTHOR model data does not exist in reality, that is if we don't have instances of Authors we can't reference them
- Since we don't have exact details related to them, we can't use the pre-defined fixtures. This is a result of lost migration tree and models. We need to keep this in mind for our development and we need to document the model structure.

gorkemarslan:
@AbhijithGanesh, I feel bad because I didn't clarify points well. 😞 

Before changing `sample_threads.json` file:

```json
    {
        "model": "api.thread",
        "fields": {
            "title": "Liquid Fluoride Thorium Reactors",
            "summary": "...",
            "category_id": "2",
            "author_id": "2",
            "created": "2016-10-09 21:07:16.848433+00"
        }
    },
```

After changing `sample_threads.json` should be like this:

```json
    {
        "model": "threads.Thread",
        "fields": {
            "title": "Liquid Fluoride Thorium Reactors",
            "summary": "...",
            "category_id": "2",
            "author_id": "1",
            "created": "2016-10-09 21:07:16.848433+00"
        }
    },
```

So it is enough to change the model name from `"api.thread"` to `"threads.thread"`. The whole process is it. No more manipulation is required.

I will want you to make one more change:
"author_id" was initially set `2`, but you should set it to `1`.

Your category.json file is completely correct. In `sample_threads.json`, just do two changes.

And don't forget that by loading these `.json` files, we add initial threads to the `CiviWiki` website. A thread must have a category and an author. `category_id` and `author_id` maps a thread to a `Category` and a `Profile` instances, respectively.

I cloned your repo again and think about why we are not on the same page the whole time. Then I found the answer with an empty database. I always assume that the initial threads will belong to the `super user`.  

As written [CONTRIBUTING.md](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/CONTRIBUTING.md):

`python manage.py createsuperuser`

I found a bug thanks to you. :) The `createsuperuser` command does not generate a `Profile`, only a `User` instance. So a super user does not have any profile. I am going to create an issue and solve it immediately.

You don't have to wait for me, of course. Just make the changes I explain to you on the top, then I approve your code. You can copy category and sample data json files from the `develop` branch and refactor via your IDE. Then register an profile on `127.0.0.1:8000/register`. When you create,

Run those commands:

```
python manage.py loaddata ./data/categories.json
python manage.py loaddata ./data/sample_threads.json
```

You'll see it is going to work in the end.

If any point is not clear to you, don't hesitate to ask me.

Thanks for you to help me find the bug.

AbhijithGanesh:
I now think both of us are on the same page and I'll use db-shell for creating a user. I'll fix it rn @gorkemarslan

AbhijithGanesh:
@gorkemarslan I have made changes, I just have one question, how fair is it to register someone else's contribution to threads as the new User's data? is it fine? I understand we don't have the data for the same too so we have to accept this.

AbhijithGanesh:
@gorkemarslan

project/data/fixtures.json
```diff
@@ -0,0 +1,1080 @@
+[
```
gorkemarslan: You should delete this file and commit again.

gorkemarslan (changes requested):

gorkemarslan:
@AbhijithGanesh, the load data has two purposes in fact:

1- Load initial data to the website to be deployed
2- Load initial data to the website to developers

Let's focus on the second.
We, as developers, and people who want to clone (download) our project need some data on the database. We want to help them to give initial data. So no one needs to create 20 threads and 20 categories thanks to `sample_threads.json` and `categories.json`.


We don't have to use the fixtures for this one:

> 1- Load initial data to the website to be deployed

If we want, we can create multiple profiles and change `author_id` fields in the fixtures to assign threads to these profiles. Then, we load json files.

I hope I can clarify the points now. 😅

AbhijithGanesh:
I understand now Gorkem Arslan, can you please approve?

gorkemarslan:
Could you delete `fixtures.json` file? We don't need it. I started a request change to remove it but I guess you missed it.

By the way, did you load json files with these commands

```
python manage.py loaddata ./data/categories.json
python manage.py loaddata ./data/sample_threads.json
```

I just want you to see the files are correct and they really work.

gorkemarslan (approved):

brylie (approved):

PR #1110, opened 2021-10-06
Use `User` model for the author field of `threads` app models (Closes #1097)

### LINKED ISSUE #1097: Change Thread.author FK to point to the `User` model instead of `Profile`
Our `Thread.author` field is currently a foreign-key relationship to the `Profile` model. However, user profiles are to some extent optional. Furthermore, the lack of default Profile object for each user has led to some uncertainty in our database fixture data.

## Task

- [ ] change the `Thread.author` model to be an FK relationship to the `User` table
- [ ] make sure related views, templates, and other functions still work as expected
- [ ] if needed, add model properties to make it easy to fech a `user.profile` and `thread.author_profile`

## Original issue

Issues #1079 and #1081 are about loading initial data to the project.

```json
    {
        "model": "threads.Thread",
        "fields": {
            "title": "Liquid Fluoride Thorium Reactors",
            "summary": "...",
            "category_id": "2",
            "author_id": "1",
            "created": "2016-10-09 21:07:16.848433+00"
        }
    },
```

and

```
python manage.py loaddata ./data/categories.json
python manage.py loaddata ./data/sample_threads.json
```

According to them, there should be at least a profile instance on the database. It is assumed to be `superuser`. 

But the `superuser` does not have any profile. So running the command below:

`python manage.py createsuperuser`

creates a user instance:

```
>>> from django.contrib.auth import get_user_model
>>> get_user_model().objects.count()
>>> 1
```

but not creates a profile:

```
>>> from accounts.models import Profile
>>> Profile.objects.count()
>>> 0
```

So in the `OpenCiviWiki`, no registered profile has superuser privileges unless given from `admin` panel. 

@brylie, if this situation is not an intended decision in the project, I want to override `createsuperuser` so that

`python manage.py createsuperuser`

creates a profile.

gorkemarslan:
Closes #1097 

As I proposed in this [post](https://github.com/CiviWiki/OpenCiviWiki/issues/1097#issuecomment-933673290), I replaced the `Profile` model with the `User` model for the `author` field.

Since this is a complicated task, this PR can remain open for at least one day to discuss whether what can be done more.

**Additional Notes:**

1. Previously, `Activity` model has the `account` field. Instead of renaming it to `profile`, I converted this model to `User` so now the name of this field is `user`.
2. From now on, after the project is initially set up and a superuser is created, the initial data can be loaded successfully by these line:
```
python manage.py loaddata ./data/categories.json
python manage.py loaddata ./data/sample_threads.json
```
3. The `Profile` model has OneToOne relation with the `User` model:
```python
class Profile(models.Model):
    user = models.OneToOneField(User, on_delete=models.CASCADE, related_name="profile")
```
4. `UserSerializer` and `UserListSerializer`are added.


project/threads/api.py
```diff
@@ -64,11 +65,11 @@ def get_thread(request, thread_id):
     try:
         t = Thread.objects.get(id=thread_id)
         civis = Civi.objects.filter(thread_id=thread_id)
-        req_a = Profile.objects.get(user=request.user)
+        requested_user = request.user
 
         # TODO: move order by to frontend or accept optional arg
         c = civis.order_by("-created")
-        c_scores = [ci.score(req_a.id) for ci in c]
+        c_scores = [ci.score(requested_user.id) for ci in c]
```
brylie: Can we just use `request.user.id` here directly?
```suggestion
        c_scores = [ci.score(request.user.id) for ci in c]

```

project/threads/api.py
```diff
@@ -64,11 +65,11 @@ def get_thread(request, thread_id):
     try:
         t = Thread.objects.get(id=thread_id)
         civis = Civi.objects.filter(thread_id=thread_id)
-        req_a = Profile.objects.get(user=request.user)
+        requested_user = request.user
```
brylie: Unless this is used elsewhere, let's use `request.user` directly in related code.
```suggestion

```

project/threads/api.py
```diff
@@ -150,14 +151,14 @@ def get_responses(request, thread_id, civi_id):
        This is used to get responses for a Civi
     """
     try:
-        req_acct = Profile.objects.get(user=request.user)
+        requested_user = request.user
```
brylie: Let's use request.user directly in related code
```suggestion

```

project/threads/api.py
```diff
@@ -150,14 +151,14 @@ def get_responses(request, thread_id, civi_id):
        This is used to get responses for a Civi
     """
     try:
-        req_acct = Profile.objects.get(user=request.user)
+        requested_user = request.user
         c_qs = Civi.objects.get(id=civi_id).responses.all()
         c_scored = []
         for res_civi in c_qs:
-            c_dict = res_civi.dict_with_score(req_acct.id)
+            c_dict = res_civi.dict_with_score(requested_user.id)
```
brylie:
```suggestion
            c_dict = res_civi.dict_with_score(request.user.id)

```

project/threads/api.py
```diff
@@ -150,14 +151,14 @@ def get_responses(request, thread_id, civi_id):
...
+        requested_user = request.user
         c_qs = Civi.objects.get(id=civi_id).responses.all()
         c_scored = []
         for res_civi in c_qs:
-            c_dict = res_civi.dict_with_score(req_acct.id)
+            c_dict = res_civi.dict_with_score(requested_user.id)
             c_rebuttal = res_civi.responses.all()
             if c_rebuttal:
-                c_dict["rebuttal"] = c_rebuttal[0].dict_with_score(req_acct.id)
+                c_dict["rebuttal"] = c_rebuttal[0].dict_with_score(requested_user.id)
```
brylie:
```suggestion
                c_dict["rebuttal"] = c_rebuttal[0].dict_with_score(request.user.id)

```

project/threads/api.py
```diff
@@ -181,10 +182,10 @@ def create_civi(request):
     :return: (200, ok) (400, missing required parameter) (500, internal error)
     """
 
-    a = Profile.objects.get(user=request.user)
+    user = request.user
```
brylie:
```suggestion

```

project/threads/api.py
```diff
@@ -206,44 +207,44 @@ def create_civi(request):
...
                 notify.send(
                     request.user,  # Actor User
-                    recipient=parent_civi.author.user,  # Target User
+                    recipient=parent_civi.author,  # Target User
                     verb=u"responded to your civi",  # Verb
                     action_object=civi,  # Action Object
                     target=civi.thread,  # Target Object
                     popup_string="{user} responded to your civi in {thread}".format(
-                        user=a.full_name, thread=civi.thread.title
+                        user=user.full_name, thread=civi.thread.title
```
brylie: I'm having a bit of trouble following the logic here. Can we just use `request.user` her? Is the code behaving as expected?
```suggestion
                        user=request.user.full_name, thread=civi.thread.title

```

project/threads/api.py
gorkemarslan: @brylie, I'll do. I was editing my PR message to write changes. Could you check it?

project/threads/api.py
```diff
@@ -206,44 +207,44 @@ def create_civi(request):
...
                         target=civi.thread,  # Target Object
                         popup_string="{user} created a new civi in the thread {thread}".format(
-                            user=a.full_name, thread=civi.thread.title
+                            user=user.full_name, thread=civi.thread.title
                         ),
                         link="/{}/{}".format("thread", thread_id),
                     )
 
-        return JsonResponse({"data": civi.dict_with_score(a.id)})
+        return JsonResponse({"data": civi.dict_with_score(user.id)})
```
brylie: It would be useful here to be explicit whether this is the `request.user`, since this code is somewhat long and convoluted.

```suggestion
        return JsonResponse({"data": civi.dict_with_score(request.user.id)})

```

project/threads/api.py
```diff
@@ -254,7 +255,7 @@ def rate_civi(request):
     """ Use this function to rate a Civi """
     civi_id = request.POST.get("civi_id", "")
     rating = request.POST.get("rating", "")
-    account = Profile.objects.get(user=request.user)
+    user = request.user
```
brylie: Similarly, let's not alias `request.user`, so downstream code is easier to understand.
```suggestion

```

project/threads/api.py
```diff
@@ -264,12 +265,12 @@ def rate_civi(request):
         )
 
     try:
-        prev_act = Activity.objects.get(civi=voted_civi, account=account)
+        prev_act = Activity.objects.get(civi=voted_civi, user=user)
```
brylie:
```suggestion
        prev_act = Activity.objects.get(civi=voted_civi, user=request.user)

```

project/threads/api.py
```diff
@@ -264,12 +265,12 @@ def rate_civi(request):
...
 
     try:
-        prev_act = Activity.objects.get(civi=voted_civi, account=account)
+        prev_act = Activity.objects.get(civi=voted_civi, user=user)
     except Activity.DoesNotExist:
         prev_act = None
 
     activity_data = {
-        "account": account,
+        "user": user,
```
brylie:
```suggestion
        "user": request.user,

```

project/threads/api.py
gorkemarslan: There were too many lines to change and follow so I intentionally tried not to make huge changes. This is the reason the code is not clear. I'll change all of them with `request.user`.

brylie (commented):
Overall this looks great, and I don't want to block it from being merged.

The only changes I would suggest are related to using the `request.user` directly in code, rather than creating aliasing variables. Using `request.user` directly would make the code more clear and consistent.

gorkemarslan:
@brylie, one more issue I observe:
If we create user and superuser on the Django shell,  `Profile` instances aren't created for them. Our `login` mechanism works to control whether a user (a `User` instance) exists. So after creating users from shell, these users are able to successfully login because they have User instances on the database. 

Please remember that after successfully login, users are redirected to `base_view` :

https://github.com/CiviWiki/OpenCiviWiki/blob/d507b0d47908d02dc40ceb0110a96d0e0f156d35/project/frontend_views/views.py#L16-L21

Obviously, these users need to have Profile instances, as well. But they don't so after login, the project is broken and raises an error like there is no such a `Profile` instance. For this reason, each user has to have a `Profile` instance. If it is OK, I am planning to override methods for creation of user and superuser so that they'll have Profile instances just after created.

brylie:
> I am planning to override methods for creation of user and superuser so that they'll have Profile instances just after created.

A `User` model `post_save` signal receiver should be all that is needed:

https://stackoverflow.com/q/63057047

brylie (approved):

gorkemarslan:
@brylie, I've recently been busy, especially for the last days, but still try to do my best.

I forgot to mentioned that:

Now, we're automatically creating `Profile` instances after creation of `User` instances thanks to the `django-signals`.

```python
        user = get_user_model().objects.create_user(username="testuser", email="test@test.com", password="password123")
```
As a result of this operation, a `Profile` instance is created.

Let's keep in mind that we should completely forbid Profile instance creation with `Profile.objects.create()` .

So, if you see that

```python
        user = get_user_model().objects.create_user(username="testuser", email="test@test.com", password="password123")
        self.test_profile = Profile.objects.create(user=user, first_name="Test", last_name="User", about_me="About Me")
```
in someone's commits in the future, please request changes to remove the bottom line. Creating `Profile` instance with these two lines can be natural for someone who is unfamiliar with `signals`. Otherwise, in the run time, an error like `there's already such a Profile instance on the database` raises.

So l just want to inform you about the possible outcomes to use signals. Let's be careful with this situation. :wink:

Issue #1129, opened 2021-10-12
[Epic] Port all Backbone code to standard Django templates


brylie:
### Idea summary

Many of our templates are defined with Backbone JavaScript code. However, we want to simplify our project and follow conventional Django development practices. To this end, we want to convert our Backbone code to Django templates.

### Further details

This task includes several sub-tasks, outlined below.

Note: this list is work-in-progress, pending some changes to the project folder structure.
- [x] #1153
- [x] #1151
- [ ] #1152

odinmay:
Interested in working on this, could you assign it to me? Ill take a look at it

brylie:
@odinmay, would you help us clarify the scope of work here? E.g., we need to list all of the Backbone templates that need to be ported over to Django template syntax. Could you help by creating that list as a comment on this issue?

odinmay:
@brylie Yeah I can look into it later today. I may need some help as well, this is one of my first contributions but I'm looking forward to trying my best to help

brylie:
Ok, sounds good. Thanks, @odinmay.

We'll learn about what code changes are needed by looking through the existing code structure. Once we identify a list of Backbone templates, we can divide and prioritize the work.

AbhijithGanesh:
Can I be assigned to this along with @odinmay to reduce workload ?

brylie:
@AbhijithGanesh, would you likevto help make the initial list of Backbone templates? Once we have the initial list of Backbone templates, the work can be divided up.

AbhijithGanesh:
I'll develop the list and update it here asap.

AbhijithGanesh:
We have about 26 templates, they're as follows:
-

-  account.html
- \accounts\templates\accounts\account.html:
- \accounts\templates\accounts\feed.html:
- \accounts\templates\accounts\password_reset.html:
- \accounts\templates\accounts\reset_by_email.html:
- \accounts\templates\accounts\settings.html:
- \accounts\templates\accounts\user-setup.html:
- \accounts\templates\accounts\partials\account\account_base.html:
- \accounts\templates\accounts\partials\account\account_settings.html:
- \accounts\templates\accounts\partials\account\login.html:
- \accounts\templates\accounts\register\entry_base.html:
- \accounts\templates\accounts\register\login.html:
- \accounts\templates\accounts\register\register.html:
- \accounts\templates\accounts\users\password_reset_complete.html:
- \accounts\templates\accounts\users\password_reset_confirm.html:
[...]

Issue #1051, opened 2021-09-28
Load Materialize CSS from a CDN


brylie:
### Idea summary

Our project currently keeps the full Materialize CSS under revision control. Since Materialize CSS is an external dependency, we don't need to keep it under revision control. Furthermore, the CSS is not minimized. To improve both of thess issues, we can load the minimized Materialize CSS directly from a CDN.

### Further details

- [x] delete the Materialize CSS files from our project
    - [x] https://github.com/CiviWiki/OpenCiviWiki/blob/524e02d9e7a46fd1c96451ecfb0d67f9fa7dc84b/project/threads/templates/static/dependencies/materialize.min.css
    - [x] https://github.com/CiviWiki/OpenCiviWiki/blob/524e02d9e7a46fd1c96451ecfb0d67f9fa7dc84b/project/threads/templates/static/dependencies/materialize.min.js
- [x] replace any Materialize CSS import with a CDN link (markup below)
    - [x] https://github.com/CiviWiki/OpenCiviWiki/blob/968a8df56212e88154e5129f305f798f787f4a40/project/threads/templates/threads/base/links.html
    - [x] or add it to our `base.html`, if issue #1049 is complete.

### Materialize CDN markup

```html
<!-- Compiled and minified CSS -->
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/css/materialize.min.css">

<!-- Compiled and minified JavaScript -->
<script src="https://cdnjs.cloudflare.com/ajax/libs/materialize/1.0.0/js/materialize.min.js"></script>         
```

AbhijithGanesh:
I want this please, I know I've been taking load but this will genuinely be useful for my learning curve! I would be extremely grateful if I am assigned to this!

AbhijithGanesh:
@brylie can I please get this!

brylie:
@AbhijithGanesh, thanks! I assigned you to the issue.

Please let us know if you have any questions or concerns.

AbhijithGanesh:
I don't have any questions  but if I don't get  my other PR merged, it 'd be hard for me to work on this due to unmerged differences @gorkemarslan and @Brylie. Is that okay?

brylie:
@AbhijithGanesh, in pull request #1048, we had to restore the `materialize.min.css` and `materialize.min.js` files so the project would run. So, we still need to delete them as part of this task.

[...]

PR #1160, opened 2021-10-21
Fix #1080 Materialize Upgrade

### LINKED ISSUE #1080: Materialize is not defined
### Description

After login,

I see a nearly empty page.

Screen Shot 2021-10-02 at 11 13 03

On the developer console of my browser, there is an error about Materialize JS:

Screen Shot 2021-10-02 at 11 14 14

> base.cb801f70fa8d.js:22 Uncaught ReferenceError: Materialize is not defined

Two recent merging requests of removing Materialize CSS and JS files (#1065 and #1067) could cause this issue.

guido97:
Closes #1080
  
## Description
Change use for modals methods, fix calls to toast, use AutoInit(),
The automatic initialization has been removed in Materialize Version 1

In general, update according to official guide and test changes
https://github.com/Dogfalo/materialize/blob/v1-dev/v1-upgrade-guide.md

### DIFF
--- a/project/accounts/templates/accounts/feed.html
+++ b/project/accounts/templates/accounts/feed.html
@@ -92,130 +92,129 @@
 
 <script id="new-thread-template" type="text/template">
     {% verbatim %}
-    <div class="modal new-thread-modal">
-        <div class="modal-content">
-            <div class="new-thread-contention new-thread-container">
-                <div class="input-field white-background-input-field left-input-field">
-                    <input id="thread-title" type="text" length="120">
-                    <label for="thread-title">Title</label>
-                </div>
-            </div>
-            <div class="new-thread-body white-background-input-field new-thread-container">
-                <div class="input-field left-input-field">
-                    <textarea id="thread-body" type="text" class="materialize-textarea" length="4000"></textarea>
-                    <label for="thread-body">Summary</label>
-                </div>
-            </div>
-    
-            <div class="new-thread-category-selection white-background-input-field" id="new-thread-category-selection">
-                <div class="input-field left-input-field">
-                    <select id="thread-category">
-                        <option value="" disabled selected>Choose your category</option>
-                        {{# _.each(this.options.categories, function (c) { }}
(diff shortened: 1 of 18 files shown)


brylie:
@guido97, I'm having difficulty reviewing these changes since there are a lot of changed lines in the template files. Are there some particular aspects of the templates that needed to change?

@gorkemarslan just merged a similar pull request. Were the changes similar to this one that fixes the Materialize errors?

guido97:
Hi @brylie, my changes are for finish update to the version 1.0.0 of materialize (Latest release), the most of these changes are for calls to toast. The version 0.97.6 is from April 1, 2016, perhaps works, but is oldest. 
In this new version automatic initialization was removed, but we can use Materialize.AutoInit (), therefore some initializations I commented. in the image below you can see modal and toast working


screenshot-modal_and_toast

gorkemarslan:
@guido97, as @brylie says, it is going to a bit difficult to follow changes in templates and JS codes.

Issue #1080 appeared after we removed `dependencies` folder and change the static Django template tags with related CDN, as can be seen in issue #1051. So our initial purpose here should be to fix the problem on the similar base and should be simple as in PR #1080, Upgrading materialize framework from` 0.97.6` to `1.0` can be considered as a further step after PR #1080. But we try to give weight to using Django templates if possible and reducing code dependencies on JS. (See similar discussion in #1129). 

I am skeptical to upgrade Materialize here. You notice upgrading from `0.100.2` to `1.0`. That's nice. However, there should be some changes from `0.97.6` to `0.100.2`. As far as I see, there is lack of tutorials for the older versions of materialize on the Internet. So upgrading can cause further side effects that can be detected harder because someone has to check all pages and JS codes to make sure there is no problem. This needs some time and energy, of course.

Keeping simple and focusing on fixing opened issues should be our strategy.

Could you check current `develop` branch work well with PR #1080? If you find new issues, could you report them by creating a new `Issue`?

gorkemarslan:
Fixed by #1158

PR #1180, opened 2021-10-30
Fix issues in threads (Closes #1170)

### LINKED ISSUE #1170: Fix errors when creating a new thread and showing existing threads
### Description

- When a new thread is created, 403 error appears in the front end.
- Some parts remain the same after `thread.author` field change from `Profile` to `User`. For example:
https://github.com/CiviWiki/OpenCiviWiki/blob/ab96113090654bc78c3674a47e46c2f9aa08d2b3/project/threads/models.py#L48-L51
- Use f-strings in `threads/models.py`
- Uploaded images should be saved under `media` folder.

gorkemarslan:
Closes #1170

# Changes

I try to solve all errors that occur when creating a new thread and a new civi. Since it requires many changes to overcome all problems, I have to make changes in various parts of the project. I'll cover all my changes, if I omit one or one is unclear please leave a comment.

- Add csrf tokens to JS files to solve 403 errors.
- Previously, we uploaded files by `PathAndRename("")`, it caused save images inside the related app without a `media` folder.

I change it in order to upload images below `media/profile_uploads`, `media/civi_uploads`, `media/thread_uploads`. 
For example:
Screen Shot 2021-10-30 at 15 48 52

Since it affects models, I apply `makemigrations`.

- I move `check_image_with_pil` and  `save_image_from_url` to `common/utils.py` because they are helper functions.
- I create a function of `check_database` inside `common/utils.py` because in the project we have:

https://github.com/CiviWiki/OpenCiviWiki/blob/ab96113090654bc78c3674a47e46c2f9aa08d2b3/project/threads/api.py#L88-L93

Unfortunately, [distinct(**fields)](https://docs.djangoproject.com/en/dev/ref/models/querysets/#distinct) is used only for `PostgreSQL` database engine. As noted in Django docs:

> On PostgreSQL only, you can pass positional arguments (*fields) in order to specify the names of fields to which the DISTINCT should apply.

We use `SQLite` in the development stage, the project is broken because `SQLite` does not have such a function like `distinct(**fields)`. It only has `distinct()` without passing parameters. I simply check whether `SQLite` database is running. According to the return value, the decision is made between` distinct() `and `distinct(**fields)`.

```python
        "contributors": [
            Profile.objects.chip_summarize(p)
            for p in Profile.objects.filter(
                pk__in=c_qs.distinct("author").values_list("author", flat=True)
            )
        ]
        if not is_sqlite_running
        else [
            Profile.objects.chip_summarize(p)
            for p in Profile.objects.filter(
                pk__in=c_qs.values_list("author", flat=True).distinct()
            )
        ],
```

It seems a duplication in code but we should not check the database for each iteration. In the future, we can optimize this part. Or we might decide to change distinct() with another function. 

- I get images from URL with `requests` lib because it is more popular than `urlib`. We had:
https://github.com/CiviWiki/OpenCiviWiki/blob/ab96113090654bc78c3674a47e46c2f9aa08d2b3/project/threads/api.py#L420

`urllib.urlretrieve` was renamed as `urllib.request.urlretrieve` and [urllib.request.urlretrieve](https://docs.python.org/3/library/urllib.request.html#legacy-interface) will be deprecated soon:

> The following functions and classes are ported from the Python 2 module urllib (as opposed to urllib2). They might become deprecated at some point in the future. Its alternative is [urllib.request.urlopen](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen).  

There are also many urrlib library: `urlib`, `urllib2`, `urllib3`. As you can see, urllib is not safe and prone to trouble.

- Unfortunately, I could not upload images to the `media` folder. The configuration for `media` seems OK to me but it did not work. So I eventually handle it with the `url()` function and it works finally. See below:
```python
urlpatterns = [
   . . .
    url(r"^media/(?P<path>.*)$", serve, {"document_root": settings.MEDIA_ROOT}),
]

urlpatterns += static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```

If you find another way or my mistake here, I am open to your suggestions.

- A `User` instance does not have `full_name` field so it must be `author.profil.full_name `not `author.full_name`.
https://github.com/CiviWiki/OpenCiviWiki/blob/ab96113090654bc78c3674a47e46c2f9aa08d2b3/project/threads/models.py#L48-L51

The solution is:
```python
"full_name": thread.author.profil.full_name, 
```

# Current Situation

After load initial data:
Screen Shot 2021-10-30 at 14 49 34

After creating a new thread with an image uploaded from link:
Screen Shot 2021-10-30 at 14 57 22

After creating a new thread with an image uploaded from local storage:
Screen Shot 2021-10-30 at 15 24 15

After creating a new civi with an image uploaded from link:
Screen Shot 2021-10-30 at 15 39 27

---
@brylie and @AbhijithGanesh, could you clone my repo and run the project to check whether there exists any error that I miss? Please report back if you find any, then I'll take care of it. This PR can be open for 2-3 days.


AbhijithGanesh:
I'll clone and review this . Thank you this is some amazing work!

brylie (approved):

PR #1188, opened 2021-11-03
Added required = true for the first name and last name form

### LINKED ISSUE #127: No verification when user deletes it's info
If I am a logged in user with properly entered information and I go to Settings and delete my first and last name, the changes are saved without the application telling me they are mandatory.
Then, if I decide to open another page, I am redirected to the registration form to complete my sign up.

khushkaushal:
Closes #127 
  
## Description
This pull request will solve the issue where users are able to save the settings form without entering first name and last name. 
made required=True to resolve the bug. All tests passed.

### DIFF
--- a/project/accounts/forms.py
+++ b/project/accounts/forms.py
@@ -127,8 +127,8 @@ class ProfileEditForm(forms.ModelForm):
             "email",
         ]
 
-    first_name = forms.CharField(label="First Name", max_length=63, required=False)
-    last_name = forms.CharField(label="Last Name", max_length=63, required=False)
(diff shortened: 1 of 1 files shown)


brylie:
Let's fix the issue in a way that allows people to remain anonymous without falsifying their first and last names. Our code should be resilient to profiles without `first_name` and `last_name` fields, which should remain optional.

brylie (changes requested):
Let's fix the issue in a way that allows people to remain anonymous without falsifying their first and last names. Our code should be resilient to profiles without first_name and last_name fields, which should remain optional.

PR #1207, opened 2021-11-21
Fix setup

### LINKED ISSUE #1206: Deprecate setup wizard
Deprecate the setup wizard.

## Task

- [ ] users should only fill out a single registration form to sign up
- [ ] users should be automatically logged in when signing up
- [ ] all "setup wizard" code should be removed from project

## Original report

I carried out the following steps:
 - Initialized an user
 - Tried accessing `My Profile`
 - Tried Setting up the user in `setup`

The issue is that, when I access the setup template view, the code breaks. This potentially could mean that we have a malfunctioning code-bit added to our main branch. 

The user should have setup the form without much hassle.

All

### Further details

image

AbhijithGanesh:
Closes #1206
  
## Description
Fixes the various Mark-up errors caused while updating from backbone templates.


AbhijithGanesh:
@gorkemarslan Can you help me with a problem?
I understand the view function passes data to our static html, but I don't understand how we can access the user model from django's template syntax. Is it fine if I pass extra detail to the frontend by adding the data related to followers, Issues and Civis?

gorkemarslan:
@AbhijithGanesh, in Django projects, user models are passed to the template with the `context processors`:

https://github.com/CiviWiki/OpenCiviWiki/blob/9a9f82fe0834e06d78f5d9ec0ee430d3b28f9749/project/core/settings.py#L82-L87

so  `"django.contrib.auth.context_processors.auth"` is responsible for this action. 

You can access a user instance in the template via `{{ user }}`. If you want to get profile image URL, you should use 
`{{ user.profile.profile_image_url }}` in the template.

However, apart from the user model, you have to explicitly pass a model (Thread, Civi, etc) to the `context` dictionary in the relevant view from the `views.py`

---
Could you explain more your purpose here? In which template you want to access these models?

> Is it fine if I pass extra detail to the frontend by adding the data related to followers, Issues and Civis?


If want to access civis of a user you can use in the templates:

```
{% for civi in user.civis %}
   <h2>{{ civi.title }}</h2>
{% endfor %}
```

Remember that there is one-to-many relation with the User model and the Civi model so we can access civis by `user.civis`

https://github.com/CiviWiki/OpenCiviWiki/blob/9a9f82fe0834e06d78f5d9ec0ee430d3b28f9749/project/threads/models.py#L186-L194

I hope I could help. If not, could you explain your purpose and your problem a bit more?

Issue #1235, opened 2022-01-10
Upgrade Python to 3.8.12


gorkemarslan:
In the project, we've specified the minimum version of Python as 3.7. Nevertheless, it should be upgraded to 3.8.12 to be able to be compatible with Django 4. After this change, we'll upgrade Django to 4, which is the last release. (PR #1234)

- [ ] Update [main.yml ](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/.github/workflows/main.yml) to have Python 3.8.12

- [ ] Update our [contributing guide](https://github.com/CiviWiki/OpenCiviWiki/blob/develop/CONTRIBUTING.md#development) to be written Python 3.8+ instead of 3.7+

ArdaGunduz00:
I will help the first open source projects, can I join?

gorkemarslan:
Hi Arda,
There is a pending PR (#1237 ) to solve this issue. I'll approve that PR in a little while.

Thank you for your interest. If you want, you can look at the other issues in the project.

ArdaGunduz00:
> Hi Arda, There is a pending PR (#1237 ) to solve this issue. I'll approve that PR in a little while.
> 
> Thank you for your interest. If you want, you can look at the other issues in the project.

Thanks.
I will look into other issues

PR #1245, opened 2022-01-26
Update django-notifications (Closes #1244)

### LINKED ISSUE #1244: Update django-notifications-hq
Since the current version of `django-notifications-hq` is not compatible with `Django 4` and the new release of this package seems not possible for now, we can adjust the `requirements` of this project so that `requirements` contains the commit number that solves compatibility with `Django 4`, rather than the release version.

gorkemarslan:
Closes #1244

I prefer to keep the current version of `django-notifications` commented.

After adjusting this package, we can update `Django 3.2.10` to `4`. For sure, we have to make small changes but they are not big issues.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -3,7 +3,8 @@ Django==3.2.10
 django-cors-headers==3.11.0
 django-extensions==3.1.5
 djangorestframework==3.13.0
-django-notifications-hq==1.6.0
+# django-notifications-hq==1.6.0
+git+https://github.com/django-notifications/django-notifications.git@0fd3f42f71c19ae75348425686a2fe68ffe04fec
 django-storages==1.12.3
 
 whitenoise==5.3.0


gorkemarslan:
@brylie, I was aiming to keep Python updated and the latest version but this seems a mistake. First, Python `3.10.0` was updated to `3.10.1` and then updated again to `3.10.2`.

What do you suggest? Should we adjust the version of Python to be  `3.9.10` or maybe `3.8.12`? Otherwise, Python will be constantly updated and we have to set the project each time.

brylie:
My preference for development environments is to specify a minimum compatible Python version (e.g. `^3.8`, as is done in [Poetry](https://python-poetry.org/docs/dependency-specification/#version-constraints). That way, developers can use whatever version of Python is compatible and available on their machine.

gorkemarslan:
I understand @brylie.

I will fix the Python version problem to reopen #1235. And I will decrease it to `3.8` in a short time. Sorry for the appearing problem. 😅

brylie:
All good :-)

Thanks for your continued stewardship of the CiviWiki project

gorkemarslan:
@brylie now I see a huge migration of Python developers who start using Poetry. In the future, we might also introduce Poetry in our project.

brylie (approved):

PR #1264, opened 2022-04-16
Account Deletion Functionality

### LINKED ISSUE #227: Account Deletion.
There seems to be no way to delete accounts as yet

## Task

- [ ] sketch and discuss the "delete my account" user interface design
- [ ] implement the design as an HTML template
- [ ] create a Django view to render the HTML template
- [ ] define a `POST` handler for the Django view that soft-deletes the user account
- [ ] ensure that users can only delete their own profile
- [ ] ensure that the user object remains but personally identifiable information (PII) is expunged

slau8:
Closes #227
  
## Description
This pull requests supports the deletion of an account from the settings page. 

### User Interface
The settings page includes a "DELETE ACCOUNT" option. When clicked, a confirmation pop-up appears. See the images below.

### `delete_user` API endpoint
On account deletion, to preserve content while removing personally identifiable information (first name, last name, username, email), PII of the account is removed from the database. The account's username is converted to `[deleted-i]` where `i` is the unique account number. This helps maintain the unique username constraint held by the database. A successful account deletion results in the user being routed to the home page.

### Testing
Backend tests were added to `test_api.py` to verify the functionality of the API call.
Screen Shot 2022-04-16 at 7 39 54 PM
Screen Shot 2022-04-16 at 7 40 03 PM


brylie (approved):

brylie:
Great work @slau8. Thanks for the contribution :-)

brylie:
@slau8, if you are available, would you like to help us re-create the app templates using only Django template syntax? There are several pages that use Backbone.js to render the templates, and we would like to build those from scratch using Django template syntax, borrowing from the Backbone code if needed.

slau8:
It was a team effort — @xuro4 designed and implemented the backend functionality and testing suite. It's our first open source contribution, so I'm glad we could help.

I'm not sure if I have time to help further with this project, but I'll keep it in mind and will let you know if anything changes.

PR #1282, opened 2022-06-22
Create the docker image

### LINKED ISSUE #1145: [FEAT]: Creation of a dockerhub image
### Idea summary

Docker Files

### Further details

I think we should bundle the whole application with docker and containerize our code base to migrate to anyother platform in case of future needs

The advantages are as follows:
-  Easier migrations to other Cloud Providers
-  Scalability Testing can be done
-  Data can be persistently saved
-  Usage of K8s to stabilize can be implemented
-  A load-balancer can be implemented

prakharrathi25:
Closes #1145 
  
## Description
This pull request creates a `Dockerfile` image for containerization of the code. It also adds a `docker-compose,yml` file which has the instructions to run the docker container and can be reused if we want to add multiple containers in the future. The instructions to run the Docker container have been added to the `README`. 

Note: The instructions inside `README` are a temp location and can be updated as required.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM python:3.8-slim-buster
+
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
+RUN mkdir /civi_app
+
+WORKDIR /civi_app
+
+ADD . /civi_app
+
+RUN apt-get update && apt-get install -y git
+
+RUN pip install --upgrade pip
+RUN pip install -r requirements/dev.txt
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,6 @@ version: '3'
 services:
   web:
     build: .
-    # command: bash -c "python project/manage.py makemigrations && python project/manage.py migrate && python project/manage.py runserver 0.0.0.0:8001"
     command: >
       bash -c "python project/manage.py migrate
       && python project/manage.py runserver 0.0.0.0:8000"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,15 @@
+
+version: '3'
+
+services:
+  web:
+    build: .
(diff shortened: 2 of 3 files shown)


prakharrathi25:
@brylie I have started working on the containerization issue and as of now the basic container is up and running. Looking to get some feedback on the code quality and method. I will implement the changes as required. You can test the code on my branch. I can provide access, if required.

Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
+RUN mkdir /civi_app
+
+WORKDIR /civi_app
+
+ADD . /civi_app
+
+RUN apt-get update && apt-get install -y git
```
brylie, line 12: Which of our dependencies requires git?

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+
+version: '3'
+
+services:
+  web:
+    build: .
+    # command: bash -c "python project/manage.py makemigrations && python project/manage.py migrate && python project/manage.py runserver 0.0.0.0:8001"
```
brylie: Ler's remove commented code, as a general practice.

prakharrathi25: Oh, sorry. My bad leaving it there.

Dockerfile
prakharrathi25: The `django-notifications` module is being installed using git. 

image

docker-compose.yml
prakharrathi25: Fixed it.

Dockerfile
brylie: OK, lets see if we can change that dependency to use a regular pip package rather than git, so our development dependencies are simpler.

prakharrathi25: I was going through the commit history and I think @gorkemarslan changed it to the git version. Maybe we should start by discussing why this was done in the first place.

brylie (approved):

PR #1296, opened 2022-08-25
Add Poetry config

### LINKED ISSUE #1295: Use Poetry for Python package management
### Idea summary

Poetry makes Python package management a bit more developer friendly than pip alone.

### Further details

In order to simplify our package management, such as updating to the latest compatible versions of all packages, we should adopt the Poetry package manager.

## Alternatives considered

Pipenv was popular for some time, but suffers from some controversy and stewardship issues.

brylie:
Closes #1295
  
## Description

Initializes a `pyproj.toml` with project configuration and dependencies. Adds documentation for installing/using Poetry for virtual environment management.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,7 @@ Make sure to search beforehand to see if the issue has been previously reported.
 The issue tracker should not be used for personal support requests. Please direct those to our [live chat](https://riot.im/app/#/room/#CiviWiki:matrix.org)
 
 ### Bug Reports
+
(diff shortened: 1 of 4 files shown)


gorkemarslan:
We need to remove the `requirements` folder and make git runners compatible with poetry. They are still using the `requirements` folder.

brylie:
> We need to remove the requirements folder and make git runners compatible with poetry. They are still using the requirements folder.

Ok, see the related issue #1297, which we can update to include the git runners task.

gorkemarslan (approved):

PR #1298, opened 2022-08-29
Fix failed pipelines

gorkemarslan:
Pipelines are failing. Let's not use a specific Python version and install the minimum requirement for our project, which is Python 3.8.


gorkemarslan:
I don't understand why linting fails even though I don't touch any Python files. I will be investigating it and (probably) add separate commits to fix each pipeline.

brylie (approved):

PR #1330, opened 2022-09-24
Use Poetry for package installation, with cache

### LINKED ISSUE #1303: Update GitHub Actions to use Poetry
### Idea summary

Our GitHub Actions should now use Poetry instead of pip

### Further details

We recently switched to using Poetry for project dependency management. However, our GitHub Actions runners still use pip.

https://github.com/CiviWiki/OpenCiviWiki/blob/f330cc63d90aa4ee5a573c8598d58d5da1282d8c/.github/workflows/main.yml#L21-L24

## Task

- [ ] the GitHub actions runner should run Poetry instead of pip
- [ ] the Action runner should install all project dependencies via Poetry

brylie:
Closes #1303

Primary changes
- update CI pipeline to use Poetry
- update pipeline versions (incl. Python)
- add pytest-djago to dev dependencies (triggered several other installations)
- lint everything with `black`
- prefix unused control variables with _ to pass bugbear lint
- remove `black` ignores since Django 4.x should now honor black style
- update recommended Python minimum version to 3.10


.github/workflows/main.yml
```diff
@@ -2,39 +2,62 @@ name: CI Workflow
...
-        max-line-length: "88"
-        exclude: "**/migrations/*.py"
-
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Set up python
+        id: setup-python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.10"
```
gorkemarslan: In your[ pyproject.toml ](https://github.com/CiviWiki/OpenCiviWiki/pull/1330/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R9)file, the Python version is defined as `^3.9`

Let's decide which Python version we will use. :) Also I wouldn't specify the exact Python version. Our pipelines constantly get broken because of the precise Python versions. You can use "^3.10" or "^3.9".

.github/workflows/main.yml
```diff
@@ -2,39 +2,62 @@ name: CI Workflow
...
+      #----------------------------------------------
+      #  -----  install & configure poetry  -----
+      #----------------------------------------------
```
gorkemarslan: You can remove the comment because the job is self-explanatory whereas your comments regarding cached dependencies can stay.

pyproject.toml
```diff
@@ -1,13 +1,3 @@
-[tool.black]
-exclude = '''
-(
-    /(
-      | migrations
-    )/
-    | apps.py
-)
-'''
-
```
gorkemarslan, line 10: It is a good idea to remove black configurations. We will configure pre-commit for our project. 👍

gorkemarslan (changes requested):

.github/workflows/main.yml
brylie: I'll try ^3.9.

I don't want developers to be pinned down too tightly in their Python version, since they might not be familiar with `pyenv`.

.github/workflows/main.yml
brylie: Done

.github/workflows/main.yml
brylie: Looks like ^3.9 worked

gorkemarslan (approved):

PR #1369, opened 2022-10-03
Added makemigrations.py file for issue 1351

### LINKED ISSUE #1351: disable auto-named migration
## Task

- [ ] create a file `project/core/management/commands/makemigrations.py` with the following content

```py
from django.core.management.base import CommandError
from django.core.management.commands.makemigrations import (
    Command as BaseCommand,
)


class Command(BaseCommand):
    def handle(self, *app_labels, name, dry_run, merge, **options):
        if name is None and not dry_run and not merge:
            raise CommandError("-n/--name is required.")
        
        super().handle(
            *app_labels,
            name=name,
            dry_run=dry_run,
            merge=merge,
            **options,
        )
```

michaeldakin:
Closes #1351
  
## Description
Created makemigrations.py file and added the desired lines requested by brylie.

### DIFF
--- a/project/core/management/commands/makemigrations.py
+++ b/project/core/management/commands/makemigrations.py
@@ -1,0 +1,18 @@
+from django.core.management.base import CommandError
+from django.core.management.commands.makemigrations import (
+    Command as BaseCommand,
+)
+
+
+class Command(BaseCommand):
+    def handle(self, *app_labels, name, dry_run, merge, **options):
+        if name is None and not dry_run and not merge:
+            raise CommandError("-n/--name is required.")
+
+        super().handle(
+            *app_labels,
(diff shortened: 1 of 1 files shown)


brylie:
@michaeldakin, thank you for opening this pull request. 

Please add a comment in the related issue to claim it so other developers can see which tasks are available. Also, please check first to see if another developer is already working on any related issue, so we avoid duplicate work.

brylie (changes requested):
@michaeldakin, thank you for opening this pull request.

Please add a comment in the related issue to claim it so other developers can see which tasks are available. Also, please check first to see if another developer is already working on any related issue, so we avoid duplicate work.

PR #1370, opened 2022-10-03
Add django debug toolbar

### LINKED ISSUE #1344: add django-debug-toolbar
## Task

- [ ] install `django-debug-toolbar` with the command `poetry add django-debug-toolbar --group dev`
- [ ] follow the [remaining installation instructions](https://django-debug-toolbar.readthedocs.io/en/latest/installation.html)

daaavo:
Closes #1344
  
## Description
In the first commit it adds django-debug-toolbar to poetry in case there's need to change the versions
The second commit follows the installation of django-debug-toolbar and it also changes the "DEBUG" constant variable to "True" in the settings.py file because it was set on False preventing the Toolbar to show up.

### DIFF
--- a/project/core/settings.py
+++ b/project/core/settings.py
@@ -7,7 +7,7 @@ Production settings file to select proper environment variables.
 import os
 
 # True if not in os.environ
-DEBUG = os.getenv("DEBUG", True)
+DEBUG = os.getenv("DEBUG", False)
 
 # defaults to second value if not found in os.environ
 DJANGO_HOST = os.getenv("DJANGO_HOST", "LOCALHOST")
--- a/project/core/settings.py
+++ b/project/core/settings.py
@@ -6,8 +6,8 @@ Production settings file to select proper environment variables.
 """
 import os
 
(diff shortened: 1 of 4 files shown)


project/core/settings.py
```diff
@@ -6,8 +6,8 @@
 """
 import os
 
-# False if not in os.environ
-DEBUG = os.getenv("DEBUG", False)
+# True if not in os.environ
+DEBUG = os.getenv("DEBUG", True)
```
brylie: For security, we want the default value for `DEBUG` to be `False`
```suggestion
# True if not in os.environ
DEBUG = os.getenv("DEBUG", False)
```

brylie (changes requested):
This PR looks good overall. Let's leave the default setting for `DEBUG` to `False`. I will follow up with a task to add developer documentation for setting environment variables.

project/core/settings.py
daaavo: kk i'll leave it as default and push up the changes

PR #1393, opened 2022-10-05
Port account.html file to Django template syntax

### LINKED ISSUE #1187: Uncaught TypeError: Cannot read properties of undefined (reading 'profile_image')
### Description

After going to `http://127.0.0.1:8000/profile/<username>/`, this error raises in the frontend:

Screen Shot 2021-11-03 at 20 27 19

Screenshot of this page:

Screen Shot 2021-11-03 at 20 30 43

### LINKED ISSUE #1199: port `account.html` to Django template syntax
The file `accounts/templates/accounts/account.html` contains several `varbatim` sections that render Backbone content. We want to replace the Backbone content with Django template syntax. 

- [ ] move all elements from `block backbone_template` to `block content`
- [ ] change the Backbone code to use Django template syntax
- [ ] Ensure that all of the following code is replaced by Django template syntax:

account.html - each verbatim section needs to be converted to Django template syntax and the view should provide necessary context data
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/8ced00f932235b1ad472dcae5ecc6d0b9525bcec/project/accounts/templates/accounts/account.html#L14-L89
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/8ced00f932235b1ad472dcae5ecc6d0b9525bcec/project/accounts/templates/accounts/account.html#L92-L102
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/8ced00f932235b1ad472dcae5ecc6d0b9525bcec/project/accounts/templates/accounts/account.html#L105-L139'
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/8ced00f932235b1ad472dcae5ecc6d0b9525bcec/project/accounts/templates/accounts/account.html#L142-L168
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/8ced00f932235b1ad472dcae5ecc6d0b9525bcec/project/accounts/templates/accounts/account.html#L171-L176
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/8ced00f932235b1ad472dcae5ecc6d0b9525bcec/project/accounts/templates/accounts/account.html#L195-L209
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/8ced00f932235b1ad472dcae5ecc6d0b9525bcec/project/accounts/templates/accounts/account.html#L224-L241
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/8ced00f932235b1ad472dcae5ecc6d0b9525bcec/project/accounts/templates/accounts/account.html#L246-L274
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/8ced00f932235b1ad472dcae5ecc6d0b9525bcec/project/accounts/templates/accounts/account.html#L281-L295
- [ ] https://github.com/CiviWiki/OpenCiviWiki/blob/8ced00f932235b1ad472dcae5ecc6d0b9525bcec/project/accounts/templates/accounts/account.html#L352-L369

### LINKED ISSUE #1396: FileError when editing user profile settings
### Description

After registering and filling in my initial profile, I attempt to visit the Settings screen via the user menu (pictured below).

image

When I click the "Settings" link, I am taken to the settings page at http://localhost:8000/settings/

The following `FileError` messages are displayed and the settings page looks somewhat broken.

image

I expected the settings page to look fairly "clean" and function correctly.

Chrome

### Further details

_No response_

gorkemarslan:
Close #1199 
Close #1187 
Close #1396 

- I find out an issue regarding objects of `User` and `Profile` models in `project/threads/api.py` and fix it. The issue blocks us to create a new civi.
- A couple of static template tags include leading space before the name of files so templates don't load properly.
- I drop the usage of BackBone.js and migrate the code to Django template syntax in account.html.
- After dropping BackBone.js, I am able to delete some js code because they are not used anymore. 
- I need to move `solutions` logic to the Thread model because we need to render solutions data, which was initially located in get_profile API view and sent it as an API response to the BackBone.js
- Fix some alignment issues in the design
- `profile.profile_image` is changed to `profile.profile_image_url` to fix #1187. We should use `profile_image_url` to get profile images because the function is able to return a placeholder image if a user doesn't have a profile image. 

After merging this branch, I feel like we take some initial steps to remove DRF functions and JS code. I should also add that I slightly change the design on the user profile page after dropping BackBone and js code. 

I think we can completely get rid of DRF but still need to keep some portion of JS code if we want the website to have a fancy look.


gorkemarslan:
The failure in the pytest pipeline is caused by one of the previously merged branches. I think either poetry or pytest is misconfigured. I will be investigating the issue.

brylie (approved):

brylie (approved):
Great work!

PR #1418, opened 2022-10-07
Fix pipeline

gorkemarslan:
- Addpytest configurations to point out where the settings.py file is located
- Run collectstatic command before tests

Tests failed because during the tests there is no static files in the project directory. This can be observed by locally deleting the staticfiles folder from the project and running pytest command

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,8 @@ jobs:
         run: poetry install --no-interaction --no-root
       - name: Run migrations
         run: poetry run python project/manage.py migrate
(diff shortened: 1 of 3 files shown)


brylie (approved):
As an aside, ensure there is an open issue for any significant task, such as fixing broken CI tests. Then, we can link the PR to the issue or possibly let other contributors take the task.

brylie:
@all-contributors please add @gorkemarslan for CI

brylie:
@all-contributors please add @gorkemarslan for infra

PR #1422, opened 2022-10-08
add poetry caching in main workflow

### LINKED ISSUE #1383: Fix CI caching
### Idea summary

Ensure our CI can use cached versions of Poetry.

### Further details

Our CI is supposed to use cache for Poetry. However, there never seems to be a cache hit, meaning Poetry is installed with each CI run.

## Task

- [ ] enable the CI pipeline to use a cached Poetry installation

didiermichel:
Closes #1383

## Description
Adding cache mechanism to main workflow, we ensure  to reuse poetry install whenever possible. This saves around 10s for each run (~ 14s -> 4s). This is not a huge time reduction but the ratio is (35% time reduction).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,6 +19,11 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: "^3.9"
+      - name: Load cached Poetry installation
+        uses: actions/cache@v2
+        with:
+          path: ~/.local  # the path depends on the OS
+          key: dotlocal-${{ runner.os }}-${{ hashFiles('.github/workflows/main.yml') }}
       - name: Install Poetry
         uses: snok/install-poetry@v1
         with:


didiermichel:
Here is a link to an execution that uses the cache : https://github.com/didiermichel/OpenCiviWiki/actions/runs/3211709378/jobs/5250068478
I ran this in my fork with the same changes then this PR + an other irrelevant commit on README.md file needed to trigger a second CI run in order to be able to use an actual cache. 

We can the that cache reloading takes 1s.
And the install poetry step that takes 3s. I didn't want to bypass it as it seems to do useful things. But the interesting thing is that is detects it is already installed and skip the installation part.

.github/workflows/main.yml
```diff
@@ -19,6 +19,11 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: "^3.9"
+      - name: Load cached Poetry installation
+        uses: actions/cache@v2
+        with:
+          path: ~/.local  # the path depends on the OS
+          key: dotlocal-${{ runner.os }}-${{ hashFiles('.github/workflows/main.yml') }}
```
brylie, line 26: Should this check for a hash of `pyproject.toml` instead of `main.yml`? I.e., wouldn't the Poetry version be defined in the `pyproject.toml`? 

I'm just not sure how this Poetry caching step should work.

brylie: Nevermind, I see that the specific Poetry version is defined later in this file.

brylie (approved):

brylie:
@all-contributors please add @didiermichel infra

Issue #1406, opened 2022-10-07
Research standards-based frameworks for web UI components


brylie:
### Idea summary

We need a standards-based (i.e., CSS, HTML, and Web Components) approach to defining UI components.

## Constraints

Any UI design system pattern library or framework adopted by this project should be:

- simple (avoid build tools and other complexities if possible)
- standards-based (use only CSS, JS, HTML, Web Components, and related standards defined by W3C/WHATWG)
- batteries-included (would ideally provide ready-made components rather than more primitive "utilities")
- decluttered (HTML templates should remain pretty clean without a lot of aesthetic descriptors)
- fully open source (no pro version upgrades or proprietary components)

Note: our design mockups use Material Design so a material design library may be desirable.

## Task

- [ ] research standards-based tools/frameworks for creating UI design systems
- [ ] locate any existing component libraries/design systems that we may use, with the ability to modify them as needed
- [ ] document/discuss your findings in this issue so we can compare the options

brylie:
@srishti666, would you like to work on this task?

srishti666:
I would love to work on this but cannot understand the task. Actually, I'm a beginner so I'm having problems.

brylie:
@srishti666, would you like to be a beta tester and find potential bugs? If so, I can create a task specifically for you.

srishti666:
No thanks, I think I'm not fit for this task

PR #1428, opened 2022-10-11
Accounts templates django syntax

### LINKED ISSUE #1151: Port `accounts` app templates to Django template syntax
We want to simplify the CiviWiki project by removing JavaScript SPA code written in Backbone.js and using Django templates instead. We will work through the Backbone templates in the `accounts` app to convert the template syntax from Backbone to Django. We want to maintain the visual styling. Below is a checklist of the relevant Backbone templates to be ported to complete this task.

Note: Backbone markup can be found between `{% verbatim %}` and `{% endverbatim %}` template tags as well as within `{% block backbone_template %}`.

## Task

Port the following templates to Django template syntax and ensure they render correctly via the relevant Django view. 

**Note:** all template baths below start within the `accounts/template/accounts/` directory, for brevity.

- [x] #1199
- [x] complete the following for `settings.html`
    - [x] create view for "Save Changes" form submission
    - [x] create view for "delete account" submission
    - [x] use vanilla JS confirm for "delete account" confirmation dialogue
- [x] convert Backbone view models to Django view/model in the following templates
    - [ ] `feed.html` - moved to `threads` app for follow-up work
    - [x] `settings.html`
- [x] ensure the following properties exist on the `User` or `Profile` model
    - [x]  `civis`
    - [x] `followers`
    - [x] `following`
    - [x] `categories`
    - [x] `threads`
    - [ ] `drafts`
- [x] convert API views to Django views
    - [x] `edit_user`
    - [x] `delete_user`
    - [x] `upload_profile_image`
    - [x] `clear_profile_image`
    - [x] `request_follow`
    - [x] `request_unfollow`
    - [x] `edit_user_categories`
- [x] simplify accounts views so there is only one `ProfileEdit` view
- [x] remove `ProfileActivationView`
- [x] remove `ProfileSetupView`
- [x] delete `user_setup.html` and any related routing logic (such as where we force the user through the setup wizard)
    - for simplicity, let's deprecate the setup wizard in favor of `update_settings.html`
- [ ] optionally, add `translate` tags for all accounts UI texts

brylie:
This pull request takes a step back, feature-wise, so we can take some steps forward. The JavaScript code is a significant blocker, so this pull request removes the blocking code/complexity.

My goal is to remove the `accounts.js` Backbone JS while preserving templates and view code. That way, we can re-implement the features as a community, possibly during Hacktoberfest :jack_o_lantern: 

Closes #1151 
 
The work can now continue to reimplement the profile sections.

## Changes

- remove mypy pre-commit due to repeated "missing dependency" failures
- port accounts templates over to Django
- port essential accounts code to Django views/model attributes
- add missing `reverse_name` properties on foreign keys
- reimplement follow/unfollow
- remove all API views, URLs, and tests

## Continuing work

The work to improve the user profile and feed views will continue in the following epic #1429


gorkemarslan:
I wouldn't delete `account.js` file completely. For example, we need this section to display tabs:

```js
    el: '#account',
    template: _.template($('#account-template').html()),
    settingsTemplate: _.template($('#settings-template').html()),
    sidebarTemplate: _.template($('#sidebar-template').html()),

    // Account Tabs Templates
    mycivisTemplate: _.template($('#my-civis-template').html()),
    followersTemplate: _.template($('#followers-template').html()),
    followingTemplate: _.template($('#following-template').html()),
    myissuesTemplate: _.template($('#my-issues-template').html()),

    // Partials
    userCardTemplate: _.template($('.user-card-template').html()),
```

I know we want to simplify the project and we are heavily dependent on JS right now. For this reason, it would be great to remove js but we don't have enough resources so if we delete the JS codes we need the help of designers to launch the new design and it will take some time.  instead of doing dramatic (removing `account.js `file in this case) changes, my suggestion is to do everything step by step.

So let's not touch any JS files. If we need, we can write and replace the new JS implementations with the old ones. I can help with that. 

I just see your new issue #1429. So we are on the same page. 😄 

Let me paraphrase myself then. We can create a new accounts JS file from scratch. Then we can deal with other logic as user follow, unfollow, settings tabs, profile edit, etc. When we are done, we will remove the old JS file. After that, we will be ready to remove .less files and start using UI frameworks like bootstrap or tailwind. (Issue #1406)

In short, you don't have to delete account.js file in this PR because it will be a dramatic change.

brylie:
This PR is in an early draft state, please give it a bit more time to shape up. Please also review the issue description for https://github.com/CiviWiki/OpenCiviWiki/issues/1151 to get an idea of what I intend to include in this pull request.

I firmly believe that Backbone JS harms the project by making it complicated to work with the code. I would instead create follow-up issues where we re-create the features using standard Django functionality while allowing the CiviWiki community, myself included, to take on those tasks. Hence, this pull request will remove some functionality while preserving relevant code for reuse when adding those features back in subsequent issues.

brylie:
Note that I am preserving parts of the UI code for use in follow-up tasks, where we rebuild the tabs and follow/unfollow feature.

project/accounts/models.py
```diff
@@ -104,6 +104,12 @@ def resize_profile_image(self):
         """
         Resizes and crops the user uploaded image and creates a thumbnail version of it
         """
+
+        # TODO: try to remove this resize_profile_image method
+        # or find a more simple way to acheive the goal(s)
+        # - less disk space?
+        # - desired shape?
+
```
gorkemarslan, line 118: I think it is a bit early to optimize  `resize_profile_image`. I mean this is obviously necessary but not a must for now. In the future, we can create a new issue to do so. The best approach would be to move it to the front-end because we don't want the back end to be busy because of image operations. We should leave image operations to browsers. Another thing is, as far as I can see there is no security check for images. (file format, image size etc.)

project/accounts/models.py
```diff
@@ -22,6 +23,55 @@ class User(AbstractUser):
...
+            my_issue_item = {
+                "thread_id": thread.id,
+                "thread_title": thread.title,
+                "category": thread.category.name,
+                "solutions": solutions,
+            }
+            issues.append(my_issue_item)
+
+        return issues
```
gorkemarslan: You can remove it because we can directly access to the thread list of a user anywhere with:
```python
thread = Thread.objects.filter(author=user)
civis = thread.civis.all()
# or
civis = Civi.objects.filter(author=user, thread__id=thread.id)
```

Alternatively, we can create a new view in threads app: `UserThreadListView` for this purpose

The `issues` method above increases the complexity. It is not easy to understand for what purpose it is used but basically, it returns all threads created by a user.

gorkemarslan (changes requested):

project/accounts/models.py
brylie: Yep. It's just a todo comment but isn't intended for this PR. I'll create an issue for it.

project/accounts/templates/accounts/partials/user_civis_list.html
```diff
@@ -0,0 +1,80 @@
+{% for civi in user.civis.all %}
```
gorkemarslan: Can you rename this file to `user_civis.html` or `user_civi_list.html`?

project/accounts/templates/accounts/partials/user_followers_list.html
```diff
@@ -0,0 +1,85 @@
+<p>
```
gorkemarslan: Can you rename this file to `user_followers.html` or `user_follower_list.html`?

project/accounts/templates/accounts/partials/user_threads_list.html
```diff
@@ -0,0 +1,47 @@
+{% for issue in user.thread %}
```
gorkemarslan: Can you rename this file to `user_threads.html `or `user_thread_list.html`?

project/accounts/models.py
```diff
@@ -22,70 +22,26 @@ class User(AbstractUser):
...
 
-    def followers(self, profile):
-        return [self.chip_summarize(follower) for follower in profile.followers.all()]
 
-    def following(self, profile):
-        return [self.chip_summarize(following) for following in profile.following.all()]
+# Image manipulation constants
+PROFILE_IMG_SIZE = (171, 171)
+PROFILE_IMG_THUMB_SIZE = (40, 40)
+WHITE_BG = (255, 255, 255)
```
gorkemarslan: We can move these constants to `settings.py` file:
```suggestion
# Image manipulation constants
PROFILE_IMG = {
    "SIZE": (171, 171)
    "THUMB_SIZE": (40, 40)
    "WHITE_BG": (255, 255, 255)
}
```

project/accounts/templates/accounts/settings.html
```diff
@@ -1,81 +1,98 @@
...
+                            {{form.about_me}}
+                        </div>
+                    </div>
+                    <div class='row'>
+                        <div class="col s24 m12">
+                            <span> {{form.categories.errors}}</span>
+                            {{form.categories.label_tag}}
+                            {{form.categories}}
+                        </div>
+                    </div>
```
gorkemarslan, line 69: I don't see any categories input field in the form:

image

project/accounts/views.py
```diff
@@ -97,104 +116,57 @@ def get_initial(self):
...
+
+    profile = get_object_or_404(Profile, user=user)
+
+    # Expunge personally identifiable data in user
+    expunged_user_data = {
+        "is_active": False,
+        "email": "",
+        "first_name": "",
+        "last_name": "",
+        "username": f"expunged-{ user.id }",
```
gorkemarslan, line 208: Then during the registration, we shouldn't allow usernames starting with `expunged-`.

Suppose I have an account with the username `expunged-123`. Later, a user with username `user.id = 123` wants to expunge their accounts. Then user.save() method will fail because there is already such a username in the DB.

Further, I should say I am not a fan of this approach. Let's say you are a well-known author on OpenCiviWiki and people really like your posts. Suddenly, you want to expunge your account. Then, I can open a new account with your user data because your username will be available after deleting your account. So I can easily imitate your profile by just creating a new account with your username. So I would keep the user with the same username as inactive in DB to protect our deleted users.

We can improve the logic of expunging user data. If a user account is deleted then how do we show threads created by this user? The author of the threads will be `expunged-123` and there will be empty fields in the first name and the last name sections in the issues. We can add a new `username` property to the Profile model, which returns `self.user.username` for active users and `expunged-123` or `deleted-user`  for the deleted accounts.

gorkemarslan:
First of all, you did a great job here. That is really a huge effort. 😄

project/accounts/permissions.py
```diff
@@ -1,19 +1,8 @@
-from rest_framework.permissions import BasePermission, SAFE_METHODS
+from rest_framework.permissions import SAFE_METHODS, BasePermission
 
 
 class IsProfileOwnerOrReadOnly(BasePermission):
     """Custom API permission to check if request user is the owner of the profile"""
 
     def has_object_permission(self, request, view, obj):
         return (request.method in SAFE_METHODS) or (obj.user == request.user)
```
gorkemarslan: You can delete this file because we don't use the `IsProfileOwnerOrReadOnly` permission class anywhere. 

You can also create a new permission class: `IsProfileCompleted`, which can be substituted for `full_profile` permission decorator. I noticed that you remove `full_profile` decorator. So my question is if a user doesn't complete their profile, are they able to create new issues or civis?. Creating a new post without completing the profile means anonymous users on the OpenCiviWiki platform.

project/accounts/views.py
```diff
@@ -4,24 +4,43 @@
...
+
+class ProfileUnfollow(LoginRequiredMixin, View):
+    def get(self, request, *args, **kwargs):
+        following_profile = Profile.objects.get(user__username=kwargs["username"])
+
+        self.request.user.profile.following.remove(following_profile)
+
+        redirect_to = reverse("profile", kwargs={"username": kwargs["username"]})
+
+        return HttpResponseRedirect(redirect_to)
```
gorkemarslan: A user should not follow or unfollow their own account. You can add an if statement to check it:

```python
if request.user.username === kwargs["username"]:
    return
```

gorkemarslan (changes requested):

project/accounts/templates/accounts/settings.html
brylie: Good catch. The "profile categories" feature rewrite will be split into a separate task so that we can experiment with the UI. I've added it to epic #1429

project/accounts/views.py
brylie: Changing the general approach to expunging user data is out-of-scope for this pull request since it was part of issue #227 and merged in pull request #1264.

We can open a new discussion to reappraise how we handle expunging user data when requesting their account to be deleted. Based on that discussion, we can open issues to improve the deletion/expungement feature.

project/accounts/views.py
brylie: This is now fixed in both the view and template.

project/accounts/views.py
gorkemarslan: I wasn't aware of the user delete implementation. We can discuss the logic later. 👍

project/accounts/models.py
```diff
@@ -149,33 +97,41 @@ def profile_image_thumb_url(self):
...
 
-    def __init__(self, *args, **kwargs):
-        super(Profile, self).__init__(*args, **kwargs)
-
     def save(self, *args, **kwargs):
         """Image crop/resize and thumbnail creation"""
 
         # New Profile image --
         if self.profile_image:
             self.resize_profile_image()
```
gorkemarslan, line 105: So whenever we create or update a profile, we try to resize the profile image no matter whether the profile image is edited or not. This image operation should be done during the registration and updating of the profile.

This might be out of the scope of this PR but we should consider it in the future.

project/accounts/templates/accounts/settings.html
```diff
@@ -1,81 +1,98 @@
...
+                            {{form.last_name}}
+                        </div>
+                    </div>
+                    <div class="row">
+                        <div class=" col s12 m6 ">
+                            <span class='error'>{{form.profile_image.errors}}</span>
+                            {{form.profile_image.label_tag}}
+                            {{form.profile_image}}
+                        </div>
+                    </div>
```
gorkemarslan, line 43: When I press the `Choose File` button, the file manager pop-up is not open.

Is it the same for you? Could you test uploading the profile image?

image

project/accounts/views.py
```diff
@@ -97,104 +124,56 @@ def get_initial(self):
...
             }
         )
         return super(SettingsView, self).get_initial()
 
 
-class ProfileActivationView(View):
-    """
-    This shows different views to the user when they are verifying
-    their account based on whether they are already verified or not.
-    """
```
gorkemarslan, line 109: Why do we remove profile activation? Then someone can register by using a fake email address and can create threads without any number limitation.  Or, voting can be manipulated by these fake accounts.

I think removing ProfileActivationView will cause severe security problems in the future.

gorkemarslan (changes requested):

project/accounts/views.py
brylie: All good. Your attention to detail is excellent. You've probably spotted a bug we can add to our backlog for a contributor to fix.

project/accounts/models.py
brylie: Good call. Let's add it to the backlog as a separate task. That way, a contributor might take it.

project/accounts/templates/accounts/settings.html
brylie: The file manager opens for me in Chromium. What browser are you using? Note that we are using the standard browser file chooser, so no custom logic is in our code.

project/accounts/views.py
brylie: I'll restore that view. I mistakingly thought it was related to the beta blocker or setup wizard; I can't remember precisely the source of my confusion.

brylie: I restored ProfileActivationView with some minor refactoring.

project/accounts/templates/accounts/settings.html
gorkemarslan: I updated my browser and it was fixed.

project/accounts/tests/test_views.py
```diff
@@ -153,50 +152,6 @@ def test_anonymous_users_are_redirected_to_login_page(self):
         )
 
 
-class ProfileActivationViewTests(TestCase):
```
gorkemarslan: Can you also restore tests related to ProfileActivation?

gorkemarslan (changes requested):

gorkemarslan:
Nicely done, @brylie! Just one comment then I will approve it. 🥳

project/accounts/tests/test_views.py
brylie: Done :heavy_check_mark:

gorkemarslan (approved):

PR #1432, opened 2022-10-13
Port threads to Django syntax

### LINKED ISSUE #1152: Port `threads` app templates to Django template syntax
We want to simplify the CiviWiki project by removing JavaScript SPA code in Backbone.js and using Django templates instead. We will work through the Backbone templates in the `threads` app to convert the template syntax from Backbone to Django. We want to maintain the visual styling. Below is a checklist of the relevant Backbone templates to be ported to complete this task.

**Note:** Backbone markup is written between `{% verbatim %}` and `{% endverbatim %}` template tags.

## Task

Port the following template from Backbone to Django view/template syntax.

- [ ] convert API views to Django views (if necessary)
  - [ ] `new_thread`
  - [ ] `get_thread`
  - [ ] `get_civi`
  - [ ] `get_civis`
  - [ ] `get_responses`
  - [ ] `create_civi`
  - [ ] `rate_civi`
  - [ ] `edit_civi`
  - [ ] `delete_civi`
  - [ ] `edit_thread`
  - [ ] `upload_civi_image`
  - [ ] `upload_thread_image`

- [ ] threads app shouldn't use Django REST framework
  - [ ] remove `threads/api.py` file

- [ ] threads app should contain only Class Based Views
  - [ ] convert Functıon Based Views to CBVs
  - [ ] convert ViewSets to CBVs

- [ ] all templates in `threads` app should
  - [ ] work without any JavaScript (Backbone.js)
  - [x] use only Django template syntax, such as `{% for solution in solutions %}`

- [ ] all templates in `threads` app should not contain any
  - [x] Backbone.js template syntax, such as `{{# _.each(solutions, function (s) { }}`
  - [x] `verbatim` tags
  - [x] `backbone_templates` block content

gorkemarslan:
Related to #1152

I will create new issues to fix creation of new civis, voting, adding response and recommended civis. I feel like there are some leftovers. And we need to remove thread.js. file. I am planning to create another issue (a similar one to #1151) to clean up.


gorkemarslan:
@brylie you can take a look at this screen record to gain insight into the final design:

74c42fc8afcf696fec57494923e1a8d2d67908f5

project/core/templates/static/less/thread.less
```diff
@@ -42,6 +42,19 @@ body {
...
+    }
+
+    .scrollbar{
+      height: 95vh;
+      overflow-y: scroll;
+    }
+
+    .force-overflow{
+      min-height: 100vh;
+    }
```
brylie: Let's auto-format this file for consistency.

project/threads/models.py
```diff
@@ -102,8 +102,8 @@ def image_url(self):  # TODO: move this to utils
     is_draft = models.BooleanField(default=True)
 
     num_views = models.IntegerField(default=0)
-    num_civis = models.IntegerField(default=0)
-    num_solutions = models.IntegerField(default=0)
+    num_civis = models.IntegerField(default=0)  # TODO To be dropped
+    num_solutions = models.IntegerField(default=0)  # TODO To be dropped
```
brylie, line 106: Do we still intend to show these counts, such as on the Thread page? If so, we should consider *when* / *how often* this count should run. E.g., should the count be a cached property on the Thread model that is only updated when Civis/Solutions are created/destroyed? Or do we want to run the `count` queries *every* time the template is rendered?

project/threads/models.py
```diff
@@ -208,6 +217,15 @@ def thread_sorted_by_score(self, civis_queryset, requested_user_id):
...
         )
 
+    def problems(self, thread_id):
+        return self.filter(thread__id=thread_id, c_type="problem")
+
+    def causes(self, thread_id):
+        return self.filter(thread__id=thread_id, c_type="cause")
+
+    def solutions(self, thread_id):
+        return self.filter(thread__id=thread_id, c_type="solution")
```
brylie: How will these methods be accessed? E.g. will it be something like `Civi.objects.filter(id=<some-id).problems()`? 

- https://docs.djangoproject.com/en/4.1/topics/db/managers/#adding-extra-manager-methods

Ideally, we would have direct properties on the Civi model instance, such as from a given civi: 
- `civi.problems`
- `civi.causes`
- `civi.solutions`

project/threads/urls/urls.py
```diff
@@ -1,15 +1,14 @@
...
+from django.urls import path
+from threads import views
+
+
+urlpatterns = [
+    path("thread/<int:thread_id>/csv/", views.civi2csv, name="civi2csv"),
+    path("thread/<int:pk>/", views.ThreadDetailView.as_view(), name="thread-detail"),
+    path("about/", views.AboutView.as_view(), name="about"),
+    path("support_us/", views.SupportUsView.as_view(), name="support_us"),
+    path("howitworks/", views.HowItWorksView.as_view(), name="how_it_works"),
```
brylie: While we're modifying these `urlpatterns`, let's make them consistent.
```suggestion
    path("support_us/", views.SupportUsView.as_view(), name="support-us"),
    path("howitworks/", views.HowItWorksView.as_view(), name="how-it-works"),
```

project/threads/urls/urls.py
```diff
@@ -1,15 +1,14 @@
...
+    path("declaration/", views.DeclarationView.as_view(), name="declaration"),
+    path("create-group/", views.create_group, name="create_group"),
```
brylie:
```suggestion
    path("create-group/", views.create_group, name="create-group"),
```

project/threads/views.py
```diff
@@ -25,6 +24,21 @@
...
+    context_object_name = "thread"
+    template_name = "thread.html"
+    login_url = "accounts_login"
+
+    def get_context_data(self, **kwargs):
+        context = super(ThreadDetailView, self).get_context_data(**kwargs)
+        context["categories"] = Category.objects.all()
+        context["problems"] = Civi.objects.problems(thread_id=context["thread"].id)
+        context["causes"] = Civi.objects.causes(thread_id=context["thread"].id)
+        context["solutions"] = Civi.objects.solutions(thread_id=context["thread"].id)
```
brylie: Rather than explicitly attach these to context, let's add them as properties directly on the `Thread` model. Then, the template or other code could read as follows.

- `thread.problems`
- `thread.causes`
- `thread.solutions`

project/threads/models.py
brylie: After further reading this PR, let's add these properties directly to the `Thread` model.

- `thread.problems`
- `thread.causes`
- `thread.solutions`

brylie:
```py
class Thread(models.Model):
    ...
    @property
    def problems(self):
        return self.civis.filter(c_type="problem")

    @property
    def causes(self):
        return self.civis.filter(c_type="cause")

    @property
    def solutions(self):
        return self.civis.filter(c_type="solution")
```

brylie (changes requested):
Overall, this PR looks good. Let's just attach the related `Civi` queries for problems, causes, and solutions directly to the `Thread` class.

Let's also discuss the performance implications of civi/solution counts on the `Thread` model and where/whether those counts will be used.

brylie:
> And we need to remove thread.js. file. I am planning to create another issue

Let's use #1152 as the parent issue for the pull request that entirely removes Backbone.js.

> I will create new issues to fix creation of new civis, voting, adding response and recommended civis.

If we add those issues with good descriptions, we can label them with `hacktoberfest` so other contributors might help.

project/threads/models.py
gorkemarslan: We shouldn't store number of civis and solutions to the database because we can directly get these stats with:

```python
num_civis = Thread.objects.get(id=thread_id).civis.count()
num_solutions = Thread.objects.get(id=thread_id).civis.solutions.count()
```

Since the aim of this task is to port thread template to Django template syntax I didn't want to do that in this PR. That is the reason why I added `# TODO To be dropped` comments to these fields, just for future reference. If I dropped them in this PR, then I have to attach makemigrations file.

These numbers show the number of civis and solutions in only one thread, not all threads. So we don't need cached property at this point. We want `count` queries whenever a user hit the thread-detail page and the corresponding template is rendered. 

Right now adding and removing civi functionalities don't work as expected. While fixing them, we. should take into account that when a civi is created or removed we should update number of civis with JavaScript. Good catch. 👍

project/threads/views.py
gorkemarslan: Good point 👍 I will fix them.

project/threads/models.py
gorkemarslan: Performance won't be a problem when we need the count of objects. It is not a huge burden on our backend. If you have some concerns about the performance we can count the number of civis in frontend. 😄  These are all early optimizations but as I said they don't affect the response time. When I work on porting account and thread templates to Django template syntax, I get an insight that we should reduce the number of lines in the project to make the project maintainable, which will make it easier for new contributors to involve in the project.

project/threads/models.py
gorkemarslan: `Civi.objects.filter(id=<some-id).problems()` is not what we want. A civi can be either a **problem** or a **cause** or a **solution**. (Please see `c_type` field.) `Civi.objects.filter(id=<some-id)` will return only one civi. A civi doesn't have problems. Instead, a civi itself can be a problem. So `Civi.objects.filter(id=<some-id)` might return one civi that is a problem civi.

What we want is to return **problem** civis or **cause** civis or **solution** civis of a thread. We can describe it as:
`Civi.objects.filter(thread__in=thread_in).problems()`. As you pointed out, it is also useful: `Thread.objects.get(id=thread_in).problems()`.  I will add these properties to the `Thread` model but not `Thread Manager` because we are interested in problems of one thread not multiple threads. On the other hand, we want to get multiple civis which are problems so we should add these properties to `CiviManager` at this time.

gorkemarslan: So these are not valid:

- civi.problems
- civi.causes
- civi.solutions

let me correct them 
- civi.c_type = problem
- civi.c_type = cause
- civi.c_type = solution

Instead of getting problems of a civi we get problem civis:
- civis.problems
- civis.causes
- civis.solutions

preferable:

- thread.civis.problems
- thread.civis.causes
- thread.civis.solutions

Let's take one more step as you mentioned:

- thread.problems
- thread.causes
- thread.solutions

project/threads/models.py
gorkemarslan: We can also drop these fields from `Civi` model because these data is stored by the `Activity` model:

```
    votes_vneg = models.IntegerField(default=0)
    votes_neg = models.IntegerField(default=0)
    votes_neutral = models.IntegerField(default=0)
    votes_pos = models.IntegerField(default=0)
    votes_vpos = models.IntegerField(default=0)
```

project/threads/models.py
gorkemarslan: I will add these properties to the `Thread` model, then I can remove them from `CiviModelManager`, right? If we do so, we can reduce the complexity. 

- thread.problem_civis()
- thread.cause_civis()
- thread.solution_civis()

project/core/templates/static/less/thread.less
gorkemarslan: Done ✅

project/threads/models.py
brylie: > On the other hand, we want to get multiple civis which are problems so we should add these properties to CiviManager at this time.

I'm not aware of a current use case for the CiviManager methods under discussion.

I think we're in agreement to move the ModelManager methods to the Thread model as properties since those are immediately useful.

brylie (approved):

PR #1459, opened 2022-11-01
Closes #1451 Added user categories to profile and settings

### LINKED ISSUE #1451: Reimplement user categories feature

This task will be to add a "categories" field to the User Profile ("settings") view).

## Task

- [ ] determine how to add a multi-select field to a Django form
- [ ] users should be able to select zero or more categories when editing their profile
- [ ] when a user updates their categories, the categories should be listed on the user profile page

TimOsahenru:
Users can now see their choiced category on their profile. but I'm having a bit of a challenge displaying the list of categories; the issue is, it just display's as a plain list instead of a clickable selective list, please I need assistance on how to execute this, I also tried using ModelMultipleChoiceField still got NO result, below is a screenshot of my code and a description of my challenge. Please I need your assistance.


project/accounts/forms.py
```diff
@@ -110,21 +110,29 @@ class ProfileEditForm(forms.ModelForm):
...
         ]
+        widgets = {"categories": forms.CheckboxSelectMultiple()}
 
     first_name = forms.CharField(label="First Name", max_length=63, required=False)
     last_name = forms.CharField(label="Last Name", max_length=63, required=False)
     about_me = forms.CharField(label="About Me", max_length=511, required=False)
     email = forms.EmailField(label="Email", disabled=True)
     username = forms.CharField(label="Username", disabled=True)
     profile_image = forms.ImageField(required=False)
+    # categories = forms.ModelMultipleChoiceField(
```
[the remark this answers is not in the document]
brylie: Just uncomment the code and leave it in. Also, use the default widget for this `ModelMultipleChoiceField`.

project/accounts/forms.py
```diff
@@ -120,19 +121,18 @@ class Meta:
...
     profile_image = forms.ImageField(required=False)
-    # categories = forms.ModelMultipleChoiceField(
-    #     queryset=Category.objects.all(),
-    #     required=False,
-    #     widget=forms.CheckboxSelectMultiple
-    # )
+    categories = forms.ModelMultipleChoiceField(
+        queryset=Category.objects.all(),
+        required=False,
+        widget=forms.CheckboxSelectMultiple(),
```
brylie, line 134: Good catch 🙂

brylie:
This looks good so far. How does it look in the UI?

TimOsahenru:
> 
same result sir, same result. I attached a screenshot.

Is it a UI issue or maybe I'm not rendering the right form choice, if the latter is the problem please how do I resolve this I've tried everything I could come up with. thanks

brylie:
Double-check the form template to see what the code looks like.

TimOsahenru:
> Double-check the form template to see what the code looks like.

Oh I did @brylie and I think the form is missing this tag
<section>
          <option></option>
<section>

I tried solving the issue but still couldn't I think it requires some UI skills(CSS) if there's a quick fix to this challenge please do let me know so I can work on this issue and have it closed for another one. Thanks

project/accounts/templates/accounts/account.html
```diff
@@ -45,6 +45,11 @@
           <div class="section-title">ABOUT ME</div>
           <div class="divider"></div>
           <div class="about-me">{{ profile.about_me }}</div>
+          <div class="divider"></div>
+          <div class="section-title">CATEGORIES</div>
+          {% for category in categories %}
+          <div class="about-me">{{ category.name }}</div>
+          {% endfor %}
```
brylie: We don't need to loop over all of the categories here, but would instead loop over the `profile.categories` and list them horizontally.

Since `div` is a block element, it causes the categories to stack. Instead of wrapping each category in its own `div` display the categories in a paragraph or span of text, separated by commas.

```html
{% for category in profile.categories %}
{{ category }},
{% endfor %}
```

Notice that the above code lists all the profile categories separated by commas. For bonus points, we can later figure out how to make it, so the last category in the list doesn't have a trailing comma.

## References

Check out the following documents for further info related to this comment.

- https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements
- https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements

brylie (changes requested):

project/accounts/views.py
```diff
@@ -178,12 +179,14 @@ class UserProfileView(LoginRequiredMixin, View):
 
     def get(self, request, username=None):
         profile = get_object_or_404(Profile, user__username=username)
+        categories = profile.categories.all()
```
TimOsahenru, line 182: Oh that's fine @brylie but, I already performed this logic at the backend or would you prefer I do it inside the HTML syntax?

project/accounts/templates/accounts/account.html
TimOsahenru: Oh that's fine @brylie but, I already performed the logic in the backend 
```
def get(self, request, username=None):
        profile = get_object_or_404(Profile, user__username=username)
        categories = profile.categories.all()
```
or will you'll prefer I do it using the Django templating syntax?

brylie: Using the `profile` object already available in the template would require less code since the categories property is already part of the `profile`.

TimOsahenru: Hello @brylie please I'm getting a server error after implementing this
here is the error message
`TypeError: 'ManyRelatedManager' object is not iterable`
any suggestions on what I'm missing out?

brylie: That error seems common enough to find an answer via Google or Stack Overflow. Let me know what search results you get and what you tried to resolve the error. It would also help to see what the exact line of code is that produces the error.

It may be worth considering that the code I suggested during the code review won't work without modification since I was going from memory.

TimOsahenru: Thanks @brylie I just pushed my code for review, I also helped fix the last trailing comma you mentioned, with the best logic I could come up with for now before we find something more permanent.

Also, I don't remember you mentioning anything about the issue I raised in the settings.html file

brylie: What is the issue with settings.html?

brylie: I'll add a comment in settings.html

project/accounts/templates/accounts/settings.html
```diff
@@ -10,83 +10,83 @@
...
+              <span> {{form.about_me.errors}}</span>
+              {{form.about_me.label_tag}}
+              {{form.about_me}}
+            </div>
+          </div>
+          <div class='row'>
+            <div class="col s12 m6">
+              {{form.categories.errors}}
+              {{form.categories.label_tag}}
+              {{form.categories}}
```
brylie, line 67: This looks correct. I'm not sure why the field isn't rendering. One thing to try would be to post this code and the form definition to a Stack Overflow question to get help from the broader community.

https://stackoverflow.com

TimOsahenru: Alright I'll do just that. but, I think it's a UI issue, I made an observation as to this form section missing out this tag

```
<select>
<option></option>
</select> 
```

could that be the issue?

brylie: The only UI here is whatever we write. Since we are using a Django Form to render UI fields, we need to figure out the correct way to render a ModelMultiSelect form field. Hence, asking on StackOverflow to see if what we are trying is correct.

TimOsahenru: thanks @brylie, I've asked on stackoverflow hoping to get answer soon.

brylie: One thing that might help in this case would be to create an empty testing project to work directly with the form fields. Just create a new Django project with a single app. Define two models in the app with one ForeignKey field. Create a form/template/view and see if the form field renders correctly. That will isolate the problem to give us an idea of how it should work.

TimOsahenru: Great thoughts @brylie. I'll do just that right away.

ikeoffiah: Can I pick this up, if this issue has not been resolved?

brylie: Sure thing. You may need to start a fresh branch, since this one has conflicts.

Issue #1446, opened 2022-10-21
Add "Followers" tab to the user profile view


brylie:
We recently removed some features from the user Profile view and now want to reimplement them using only Django.

This task will be to add a "Followers" tab to the user profile view. The tab should link to a sub-path of the user profile so it is clear to the viewer that they are viewing followers related to the user.

## Task
All of these tasks should be done in the `accounts` app.

- [ ] create a Django view called `UserFollowers` (in `accounts/views.py`)
- [ ] create a URL definition `profile/<str:username>/followers` that renders the `UserFollowers` view ( in `accounts/urls.py`)
- [ ] ensure the `UserFollowers` view context contains a list of users that are following the relevant user (given via the `username` in the URL) 
    - note, it should be possible to get the followers via `user.profile.followers` or just `profile.followers`
- [ ] create a template to render the user followers - for now, extending `base.html` so we have consistent branding  (in `accounts/templates/accounts/user_civis.html`)

BrunoFNRodrigues:
I'd like to work on this issue.

Nehemiah60:
I had requested to work on the Issues tab. I have realized that I need to complete, the followers tab, following, and then on the issues. Can, I start working on this, please?

Nehemiah60:
If these changes were merged I am still unable o see, the followers tab here with me. Maybe may you clarify on this, please? I am trying to understand this a little bit

Nehemiah60:
Do you have a minute, please? I can create a 5-minute video call. Let me know if it's possible, I can share a video link. Thanks.

PR #1463, opened 2022-11-05
Add Following tab to the user profile view

### LINKED ISSUE #1447: Add "Following" tab to the user profile view

This task will be to add a "Following" tab to the user profile view. The tab should link to a sub-path of the user profile, so it is clear to the viewer that they are viewing users that a particular user is following.

## Task

- [ ] create a Django view called `UserFollowing` (in `accounts/views.py`)
- [ ] create a URL definition `profile/<str:username>/following` that renders the `UserFollowing` view ( in `accounts/urls.py`)
- [ ] ensure the `UserFollowing` view context contains a list of users that the relevant user is following (given via the `username` in the URL) 
    - note, it should be possible to get profiles that the user follower via `user.profile.following` or just `profile.following`
- [ ] create a template to render the followed usernames - for now, extending `base.html` so we have consistent branding  (in `accounts/templates/accounts/user_civis.html`)

wippo-devops:
Closes #1447
  All of these tasks should be done in the accounts app.

## Description
Created a new view that shows followers of user (/profile/<user>/following). View shows clicable image icons of profile pictures of followers. If there are zero followers it shows message that no one is following.

I was thinking it might be good idea to add this to profile below the "About me" section, but it was not mentioned in the issue.


project/accounts/templates/accounts/user_civis.html
```diff
@@ -0,0 +1,41 @@
...
+      <div class="user-chip chip white">
+        <div class="user-chip-contents">
+          <a href="/profile/{{ following.user.username }}/">
+            <img src="{{ following.profile_image_url }}" alt="{{ following.user.username }}">
+            <!--<div class="name">
+                {{ following.first_name}} {{ following.last_name }}
+            </div>
+            <div class="light-gray-text atname">
+                @{{ following.user.username }}
+            </div>-->
```
brylie: Remove commented-out code

project/accounts/views.py
```diff
@@ -188,6 +188,21 @@ def get(self, request, username=None):
...
 
+class UserFollowing(LoginRequiredMixin, View):
+    """A view that shows list of users following user with given username"""
+
+    def get(self, request, username=None):
+        profile = get_object_or_404(Profile, user__username=username)
+
+        return TemplateResponse(
+            request,
+            "user_civis.html",
```
brylie: It would be more accurate for this template to be called `profile_followers.html`.

project/accounts/templates/accounts/user_civis.html
```diff
@@ -0,0 +1,41 @@
...
+          </a>
+        </div>
+      </div>
+    </div>
+  {% empty %}
+    <div class="section no-state">
+      <div class="container">
+        <div class="section">
+          <div class="center title-lato grey-text">
+            Not following any users
```
brylie: Add `i18n` tags
```suggestion
            {% translate "Not following any users" %}
```

project/accounts/templates/accounts/user_civis.html
```diff
@@ -0,0 +1,41 @@
+{% extends "base.html" %}
+
```
brylie: Load i18n for translation support.

```suggestion
{% load i18n %}
```

project/accounts/templates/accounts/user_civis.html
```diff
@@ -0,0 +1,41 @@
...
+{% block extra-css%}{% endblock extra-css %}
+
+{% block content %}
+  <div class="following-total">
+    <div>
+      {{ profile.following.count }} Following
```
brylie:
```suggestion
{% blocktranslate %}
      {{ profile.following.count }} following
{% endblocktranslate %}
```

project/accounts/templates/accounts/user_civis.html
```diff
@@ -0,0 +1,41 @@
...
+            Not following any users
+          </div>
+        </div>
+      </div>
+    </div>
+  {% endfor %}
+
+{% endblock content %}
+
+{% block extra-js %}{% endblock extra-js %}
```
brylie: We don't need this empty block.

```suggestion
```

project/accounts/templates/accounts/user_civis.html
```diff
@@ -0,0 +1,41 @@
+{% extends "base.html" %}
+
+{% block extra-css%}{% endblock extra-css %}
```
brylie: We don't need the empty block.
```suggestion
```

brylie (changes requested):

project/accounts/views.py
```diff
@@ -188,6 +188,21 @@ def get(self, request, username=None):
         )
 
 
+class UserFollowing(LoginRequiredMixin, View):
```
brylie: Two things are getting confused here, followers and following. Followers have subscribed to a user, while a given user can follow other users. Is the goal of this pull request to show:

A) profiles who are following this profile (followers), or
B) other profiles the current profile follows (following)?

wippo-devops: From my understanding of the issue #1463(Edit: Original issue was #1447) I would argue that it is B). For that reason I used the `profile.following` to get to the list of profile that are following the 'current' profile, or to be more specific, profile defined in the url call of `profile/<username>/following`.

Side-note: Looking at the open issues there also is issue #1446. From my point of view, this issue was created to solve the A).

brylie: Ok, I see that this PR follows the task definition quite closely. Sorry for my confusion.

For consistency, let's rename this view to `ProfileFollowing`.

project/accounts/views.py
```diff
@@ -188,6 +188,21 @@ def get(self, request, username=None):
...
+    """A view that shows list of users following user with given username"""
+
+    def get(self, request, username=None):
+        profile = get_object_or_404(Profile, user__username=username)
+
+        return TemplateResponse(
+            request,
+            "profile_followers.html",
```
brylie: This template name should be consistent with the view and URL definition, which both use "profile following" and not "followers."

brylie (changes requested):

project/accounts/views.py
```diff
@@ -188,6 +188,21 @@ def get(self, request, username=None):
         )
 
 
+class ProfileFollowing(LoginRequiredMixin, View):
+    """A view that shows list of users following user with given username"""
```
brylie: This docstring seems incorrect based on the issue description and the template code.

Which of the following is this pull request working with?

a) followers - people who are following a profile
b) following - people a profile follows

Please make sure we consistently use the words "following" and "followers" since they mean two things.

brylie (changes requested):
Overall, this PR looks good. There is just a bit more ambiguity in how we use "following" and "followers" here that we need to iron out.

project/accounts/views.py
wippo-devops: Agreed, I mixed this one up. Going to resolve it in the next commit.

brylie (approved):

brylie:
@all-contributors please add @wippo-devops for code

PR #1472, opened 2022-11-13
Added Unit tests for ProfileFollowing View

### LINKED ISSUE #1447: Add "Following" tab to the user profile view

belalmn-cmu:
Closes #1447

## Description
This PR aims to supply the existing (completed) Issue #1447 with some Unit Tests to keep consistent with the project's QA methods. I created a new entry in the test_views.py file, and simulated adding two profiles to a user's following list, and checking if the template includes their usernames as intended. There are also additional misc tests such as checking if the correct template "profile_following.html" is used in the view.

### DIFF
--- a/project/accounts/tests/test_views.py
+++ b/project/accounts/tests/test_views.py
@@ -215,3 +215,30 @@ class UserProfileView(BaseTestCase):
         self.assertEqual(response.status_code, 200)
         self.assertContains(response, self.user.username)
         self.assertTemplateUsed(response, "account.html")
+
+
+class ProfileFollowing(BaseTestCase):
+    """A class to test user profiles following view"""
+
+    def setUp(self) -> None:
+        super(ProfileFollowing, self).setUp()
+
+        self.user2 = get_user_model().objects.create_user(
+            username="newuser2", email="test@test.com", password="password123"
+        )
+        self.user3 = get_user_model().objects.create_user(
+            username="newuser3", email="test@test.com", password="password123"
+        )
+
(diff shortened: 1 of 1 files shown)


project/accounts/tests/test_views.py
```diff
@@ -215,3 +215,30 @@ def test_get_user_profile(self):
...
+    """A class to test user profiles following view"""
+
+    def setUp(self) -> None:
+        super(ProfileFollowing, self).setUp()
+
+        self.user2 = get_user_model().objects.create_user(
+            username="newuser2", email="test@test.com", password="password123"
+        )
+        self.user3 = get_user_model().objects.create_user(
+            username="newuser3", email="test@test.com", password="password123"
```
brylie, line 230: Should these users have distinct email addresses?

brylie (changes requested):

PR #1491, opened 2023-01-28
Added an Issues Tab

### LINKED ISSUE #1448: Add "Issues" tab to the user profile view

This task will be to add an "Issues" tab to the user profile view. The tab should link to a sub-path of the user profile so it is clear to the viewer that they are viewing Issues related to the user.

## Task

- [ ] create a Django view called `UserIssues` (in `accounts/views.py`)
- [ ] create a URL definition `profile/<str:username>/issues` that renders the `UserIssues` view ( in `accounts/urls.py`)
- [ ] ensure the `UserIssues` view context contains a list of users that are following the relevant user (given via the `username` in the URL) 
    - note, it should be possible to get the issues via `user.issue_civis`
- [ ] create a template to render the user issues - for now, extending `base.html` so we have consistent branding  (in `accounts/templates/accounts/user_civis.html`) 
    - unless the `profile_base.html` is available, in which case use that instead

Nehemiah60:
Closes #1448 
  
## Description
Creates an issues tab to the user profile.

### DIFF
--- a/project/accounts/templates/accounts/account.html
+++ b/project/accounts/templates/accounts/account.html
@@ -13,17 +13,10 @@
   <div id="sidebar-template">
     <div class="profile-image">
       <div class="big-image-cropper z-depth-1">
-        <img
-          class="responsive-img"
-          src={{ profile.profile_image_url }}
-          
-
-        >
+        <img class="responsive-img" src={{ profile.profile_image_url }}>
       </div>
     </div>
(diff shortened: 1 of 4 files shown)


brylie (approved):
Great work! :-)

brylie:
@all-contributors please add @Nehemiah60 for code

PR #1496, opened 2023-02-06
Added a factory file for the Category Model

### LINKED ISSUE #1350: add development fixtures
Model factories can help us scaffold content quickly for development purposes.

Let's define a `factory.py` file in every app containing a `models.py`. The `factory.py` files will contain factories for the models in the same app. E.g.,

- project/categories/models.py - model definitions for categories app
- project/categories/factory.py - factory definitions for categories app

## Task

- [ ] add `factory_boy` to this project with `poetry add factory_boy --group dev`
- [ ] define the [initial factory_boy factories](https://factoryboy.readthedocs.io/en/stable/index.html?highlight=install#defining-factories)

Nehemiah60:
Closes #1350
  
## Description
Created a factory file for the Category model.

### DIFF
--- a/project/accounts/factory.py
+++ b/project/accounts/factory.py
@@ -20,7 +20,7 @@ class ProfileFactory(factory.django.DjangoModelFactory):
     first_name = factory.Faker("lorem")
     last_name = factory.Faker("lorem")
     about_me = factory.Faker("lorem")
-    is_verified = factory.fuzzy.FuzzyChoice(choices=[False, False, False, False])
+    is_verified = factory.fuzzy.FuzzyChoice(choices=[False, False])
     profile_image = factory.Faker("image_url")
     profile_image_thumb = factory.Faker("image_url")
 
--- a/project/accounts/factory.py
+++ b/project/accounts/factory.py
@@ -24,13 +24,13 @@ class ProfileFactory(factory.django.DjangoModelFactory):
     profile_image = factory.Faker("image_url")
     profile_image_thumb = factory.Faker("image_url")
 
-    @factory.post_generation
-    def categories(self, create, extracted, **kwargs):
-        if not create:
-            return
-        if extracted:
-            for category in extracted:
-                self.categories.add(category)
+    # @factory.post_generation
+    # def categories(self, create, extracted, **kwargs):
+    #     if not create:
+    #         return
+    #     if extracted:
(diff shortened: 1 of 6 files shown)


project/categories/factory.py
```diff
@@ -0,0 +1,15 @@
+import factory
+
+from .models import Category
+
+
+class CategoryFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = Category
+
+    name = factory.Faker("name")
```
brylie: The ´name´factory is used to generate names or people. Let's use Faker factory ´lorem´ here, since it is more like a keyword.

https://faker.readthedocs.io/en/master/providers/faker.providers.lorem.html

project/categories/factory.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+categories = CategoryFactory.create_batch(10)
+for category in categories:
+    print(category.name)
```
brylie: I don't think we want to add these lines, since we will want to generate the fake content via a manual command.
```suggestion
```

project/categories/factory.py
brylie:
```suggestion
    name = factory.Faker("lorem")
```

brylie (changes requested):

project/categories/factory.py
Nehemiah60: Thanks, @brylie. I will work on it. If you get time, please you can also review the other PR, I mentioned. Thanks.

project/accounts/factory.py
```diff
@@ -20,7 +20,7 @@ class Meta:
     first_name = factory.Faker("lorem")
     last_name = factory.Faker("lorem")
     about_me = factory.Faker("lorem")
-    is_verified = factory.fuzzy.FuzzyChoice(choices=[False, False, False, False])
+    is_verified = factory.fuzzy.FuzzyChoice(choices=[False, False])
```
brylie: Should this also allow `True`?
```suggestion
    is_verified = factory.fuzzy.FuzzyChoice(choices=[True, False])
```

Nehemiah60: I was not so sure. Because the field had False as a default value. But I will make the changes @brylie

brylie (approved):

